PR #2, opened 2018-12-03
Experimenting with using `react-popper-tooltip` for tooltips

tmeasday:
@domyen since we wrote the `WithTooltip` component (which uses popper directly), this library has appeared: https://www.npmjs.com/package/react-popper-tooltip

The library handles the low-level tooltip stuff like `trigger` (what we called `mode`, i.e. `click` vs `hover` etc).

It ships with some styles, but not the styles we want.

I'm thinking we refactor `WithTooltip` to:
 (a) use `react-popper-tooltip` behind the scenes
 (b) change the API to better reflect the underlying API (e.g. `mode` => `trigger`, `startOpen` => `defaultTooltipShown`)

What do you think?

Also do you think we should publish our styled-component-ed `WithTooltip` as a package on npm?

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
     "react-lazyload": "^2.3.0",
     "react-popper": "^1.3.0",
     "react-popper-tooltip": "^2.7.0",
+    "recompose": "^0.30.0",
     "storybook-chromatic": "^1.2.0",
     "styled-components": "^4.1.1"
   },
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
     "react-helmet": "^5.2.0",
     "react-lazyload": "^2.3.0",
     "react-popper": "^1.3.0",
+    "react-popper-tooltip": "^2.7.0",
     "storybook-chromatic": "^1.2.0",
     "styled-components": "^4.1.1"
   },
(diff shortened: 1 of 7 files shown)


domyen:
I agree with using react-popper-tooltip, seems like it does what we need it to. 

Let's also publish WithTooltip on npm when this work is all done so other folks can use it too!

tmeasday:
@domyen this is working well I think.

The only thing yet left unimplemented is the `closeOnClick` prop -- which was always a bit of a hack tbh. Do we need it? Remember there is a `onHide` prop passed to the tooltip component that can be used to close the TT.

mohsinulhaq:
Hi guys, I am the maintainer of this library. Thanks for using it. Do let me know if you feel it requires any changes or improvements 🙂

domyen:
Thank you so much Mohsin. Will do. And I’ll shout you out when we launch
this. 🙏

tmeasday:
Hey @mohsinulhaq thanks for getting in touch and thanks for the library!

The one feature we had in our previous version of tooltips (using our own implementation of tooltips built on popper) was the ability to set a `closeOnClick` prop -- which meant in `trigger: 'click'` mode, the tooltip would close if you clicked anywhere inside the tooltip itself (in addition to the default behaviour, which is to close if you click anywhere *outside* of the tooltip). I've not looked closely at your implementation, but would that be something that would make sense to add? 

I'm not sure how urgent it is to use right now (@domyen) but it is something we've needed before, and I imagine is a fairly common use case. If it is something that makes sense for the lib to add, I'm happy to send a PR if we need it. Let me know your thoughts.

Issue #10, opened 2018-12-13
Get contributor count from GitHub


domyen:
<img width="496" alt="screenshot 2018-12-13 18 25 42" src="https://user-images.githubusercontent.com/263385/49973820-9506c100-ff04-11e8-8699-e6f7ee92d943.png">

**Behavior**
- Get contributor count from GitHub asynchronously

[Relevant Storybook local URL](http://localhost:6006/?selectedKind=screens%2FCommunityScreen%2FCommunityHero&selectedStory=not%20subscribed&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel)

Keraito:
I tried my hand on this issue, but there's a limitation with the information that the Github API provides. 

The [contributors endpoint](https://developer.github.com/v3/repos/#list-contributors) only provides me 30 contributors (ordered on amount of activity) and the [repo stats endpoint](https://developer.github.com/v3/repos/statistics/#get-contributors-list-with-additions-deletions-and-commit-counts) provides me with 100 contributors, similar to what the [contributors page on gh shows](https://github.com/storybooks/storybook/graphs/contributors). I found [something on the GH forum](https://github.community/t5/GitHub-API-Development-and/API-Github-list-contributors/td-p/9631) that it might not be possible to replicate the number shown on a repo page which is quite unfortunate. 

Did someone else do this before and maybe know how to circumvent this issue through the API?

domyen:
Thanks for trying this out @Keraito. I just tried querying the GitHub API using [this trick](https://stackoverflow.com/questions/44347339/github-api-how-efficiently-get-the-total-contributors-amount-per-repository). 

`curl -I "https://api.github.com/repos/storybooks/storybook/contributors?per_page=1"`

Where you ask for 1 contributor per page and count the number of pages. But it didn't give me the accurate count (maxed out at 399).

`
Link: <https://api.github.com/repositories/54173593/contributors?per_page=1&page=2>; rel="next", <https://api.github.com/repositories/54173593/contributors?per_page=1&page=399>; rel="last"
`

There may be API limitations for particularly expansive projects like Storybook. If so, we may have to hardcode this number. Curious if other folks have experience doing this with the GitHub API. 🤔

lucas-carl:
@Keraito @domyen It turns out, that you have to add `anon=true` to the query, because github makes users over a count of 500 anonymous.
[...]

PR #18, opened 2018-12-31
10 get contributor count from GitHub

dakebl:
Raised to suggest an answer to issue https://github.com/storybooks/frontpage/issues/10


tmeasday:
Seems like a reasonable approach given the limitations discussed in #10

PR #23, opened 2019-01-28
Retrieve and render Storybook Medium posts

### LINKED ISSUE #15: Get latest blog posts from Medium
Retrieve the latest blog posts from [Storybook's Medium](https://medium.com/storybookjs).

This package might help:
https://github.com/nikolaevra/medium-get-latest-posts

Keraito:
Fixes #15.
Apologies that this took so long.

This solution uses the `gatsby-source-medium` plugin as suggested in https://github.com/storybooks/frontpage/issues/15#issuecomment-456590716. With the gatsby server running it works, but in the Storybook it displays a `Loading(StaticQuery)` for me.


domyen:
Thanks @Keraito, a pattern we use for StaticQuery is to have pure presentational components in Storybook and hydrate them at the page level in Gatsby. In this case I'd suggest: 

1. Update `Footer` props to account for an object like `mediumPosts`
2. In `Footer.stories.js` use a mock `mediumPosts` to render the `Footer`
2. On the `___Screen.js` where `Footer` is used append the GraphQL query.

For example, here's an example of how we did this for [fetching data from GitHub](https://github.com/storybooks/frontpage/commit/afcdec4525123815af3f838f8383bbfdb0ab9357).

What do you think? 

(Also looping in our resident GQL expert @tmeasday for review)

Keraito:
Processed your feedback @domyen. Have to admit, it became quite repetitive as the Footer is used in all screen components. Is it an idea to maybe at least abstract such grapql queries to another place too? Because it's basically just the same query copy pasted into all the different places (same goes for the github contributor count)

tmeasday:
Great stuff @Keraito. I agree, let's make a HoC `withFooterData` or something?

```js
const withFooterData = Component => props => {
   return (
     <StaticQuery
       query={graphql`
         query AddonsScreenQuery {
           allMediumPost(sort: { fields: [createdAt], order: DESC }, limit: 3) {
             edges {
               node {
                 id
                 title
                 virtuals {
                   subtitle
                 }
                 medium_id
                 uniqueSlug
               }
             }
           }
         }
       `}
       render={data => <Component data={data} {...props} />}
     />
   );
 }
```

What do you think?

Keraito:
@tmeasday in the case that medium posts are the only data being queried through graphql that approach is decent. but there are a few cases where the medium query had to be combined with the github contributor query (https://github.com/storybooks/frontpage/pull/23/files#diff-42d93a1d8c39c023cf94708c926b1f00 f.e.). In that case, it doesn't make sense to separate both queries into an HOC. My idea was to only abstract the queries away since those are exactly the same everywhere, so it would become something like:

```js
export default function ___Screen(props) {
  return (
    <StaticQuery
      query={graphql`
        query ___ScreenQuery {
          ${githubRepoQuery}
          ${mediumPostsQuery}
        }`
      }
      render={data => <Pure___Screen data={data} {...props}/>}
  );
}
```

WDYT?

tmeasday:
@dakebl I would be OK with that. Can we mock it in storybook?

dakebl:
@tmeasday Yeah, we just need to export both the StaticQuery and the component itself: https://github.com/gatsbyjs/gatsby/blob/master/docs/docs/testing-components-with-graphql.md#testing-staticquery

tmeasday:
@dakebl - yes I have seen that before but the downside of that technique is that you can write stories for `Footer` now (using `PureFooter`) but you can no longer write stories for any component that itself uses `Footer`.

I think to bypass this problem we'd need to provide a `context` to the `Footer` (not `PureFooter`) that looks something like what `StaticQuery` expects. I think I've looked into this before without immediate success, hopefully you can have more luck!

domyen:
Bumping this PR 👋 

We're in the final stages of Storybook 5.0 QA so this site will be launching soon. I want to make sure to get this in there!

Keraito:
I will try to pick this up again this week

Keraito:
I couldn't quite find the time for this issue in the past week. If this is a matter of turning it into Fragments, I can provide a patch quite quickly. But if we want to resort back to using the `StaticQuery` that @dakebl 	describes, I'm not sure if I will be able to tackle the issues described by @tmeasday. So if the latter solution is prefered, it might be better for someone else to pick this up. cc @domyen

domyen:
Thanks @Keraito, everything is 👍 from my end. 

The requirements from me are to:
✅Get posts from Medium
✅Get stories in Storybook

@tmeasday do you want to do optimizations or should we go ahead and merge?

tmeasday:
@domyen Let's merge this as is. In the future I think we pull the GraphQL fragment out at the least.

domyen:
This is failing to build so I can't merge it.
```
FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory
```
@Keraito can you take a look? It first appeared [here](https://github.com/storybooks/frontpage/pull/23/commits/fa0b6918d6a75d2a9c5ada122ed6c5e3cbf1c822) 🙏

tmeasday:
I can't work this one out. Somehow this commit: https://github.com/storybooks/frontpage/pull/23/commits/fa0b6918d6a75d2a9c5ada122ed6c5e3cbf1c822 made the gatsby build fail.

tmeasday:
Gahh this was a really obvious problem in the end

PR #30, opened 2019-03-07
Edit wording of UX benefit to read better

Glazy:
Very simple commit but without either the word 'your' or 'the' before UI
it didn't read as nicely


domyen:
Good call @Glazy!

PR #32, opened 2019-04-11
Add Root Attribute addon

le0pard:
This addon help to change html or body attribute in stories (can help, for example, if you are using `:root` CSS variables for themes)

### DIFF
--- a/src/components/screens/AddonScreen/AddonScreen.js
+++ b/src/components/screens/AddonScreen/AddonScreen.js
@@ -290,6 +290,11 @@ export function PureAddonScreen({ data: { allMediumPost }, ...props }) {
           desc="Select between JSS themes."
           addonUrl="https://github.com/vertexbz/storybook-addon-jss-theme"
         />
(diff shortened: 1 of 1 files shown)


src/components/screens/AddonScreen/AddonScreen.js
```diff
@@ -290,6 +290,11 @@ export function PureAddonScreen({ data: { allMediumPost }, ...props }) {
           desc="Select between JSS themes."
           addonUrl="https://github.com/vertexbz/storybook-addon-jss-theme"
         />
+        <AddonItem
+          title="Root Attribute"
+          desc="Storybook addon, which provide ability to change html or body attribute."
```
domyen: I'd suggest simplifying the desc text –it's listed as an addon so calling it a "Storybook addon" is not needed. How about this:
```suggestion
          desc="Provides the ability to change the html or body attribute."
```

le0pard:
@domyen thanks

PR #38, opened 2019-05-01
Addon storybook-addon-i18 added to list of addons

goooseman:
Toggle the locale and directly see the result in the preview. Intl library agnostic - can be used with any intl library. Supports automatic lrt/rtl change.


domyen:
Thanks @goooseman!

PR #42, opened 2019-05-17
Use SB design system 

domyen:
**What I did**
Use Storybook's design system for common components 

**What changed visually**
Nothing should have changed visually. This is a swap-in-out job.

Note: I reorganized the component hierarchy so Chromatic is identifying the stories as "new" which requires approval.


kylesuss:
@domyen 1 thing I was unable to get to here is consolidating the `global` styles seen here:

https://github.com/storybookjs/frontpage/blob/design-system/src/components/basics/shared/global.js

I realized as well that I didn't get to this in learnstorybook.com either. Must have just slipped by somehow. Ultimately, there is a problem with the design system [loading the fonts this way](https://github.com/storybookjs/design-system/blob/9bcfa744617c4fe59d001db956cbc3139ec88aca/src/components/shared/global.js#L115). The `GlobalStyle` is re-rendered on each page change, so (1) there is a network request for the font every page change which is bad and (2) there is a flicker of the text that uses the font each time it loads. Therefore, we will have to update the design system to accommodate this quirk.

Leaving the `GlobalStyle` as-is here for now does not have that same issue.

domyen:
Thanks for the heads up. Let's skip it for now :D

kylesuss:
@domyen ticket added: https://github.com/storybookjs/design-system/issues/54

domyen:
QA'ed! This looks good to go.

PR #46, opened 2019-05-21
Add JetBrains to users list

Hypnosphi:
We use Storybook to develop and document our shared web UI components library

Unfortunately, I haven't found a place for a link to our live Storybook:
https://jetbrains.github.io/ring-ui


domyen:
Thanks! Will bring back the community gallery as part of #31 👍

Issue #25, opened 2019-01-31
Refactor all images to use gatsby-image


isaac-martin:
No images are using gatsby image - this will help with performance in #12 
Docs are here. https://www.gatsbyjs.org/packages/gatsby-image/

i'm opening this but probably won't have time to look at it for the next little while.

domyen:
Sweet! I'm definitely stoked to see how `gatsby-image` works when implemented well. 

I ran into an early hiccup trying it out (probably my own fault) so decided to wait for someone with more gatsby experience to come along. 🙏

isaac-martin:
Cool - let me see if i can get a half finished PR request implementing it on a couple of our images. That will open it up for wider implementation for everyone else who can dig in!

dakebl:
Any update on this @isaac-martin?

PR #47, opened 2019-05-21
Remove dupe assets in /static/image

domyen:
Some of our SVG assets ended up duplicated on our path to using direct imports instead of relative paths as recommended by Gatsby. This PR removes _some_ of the duplicated assets to hopefully clear up some confusion about where to put stuff. 

[See the initial commit by @dakebl](https://github.com/storybooks/frontpage/commit/1b0a536f641b384818866924e440e2dc834e7ff4) 

@dakebl do you think we can/should move the pngs/jpegs to direct imports as well?

Thanks for pointing it out @Hypnosphi!


dakebl:
Ah, nice one @domyen @Hypnosphi! Forgot to clean up after myself 😄 

I don't think we should directly import the jpgs or pngs, as I think we can take advantage of [Gatsby Image](https://www.gatsbyjs.org/packages/gatsby-image) to optimise our image quality/format. It uses   [sharp](https://github.com/lovell/sharp) under the hood to do its image processing.

I think @isaac-martin was looking at this as part of #25, but happy to pick it up if needed.

PR #53, opened 2019-07-03
Disable remote playback to hide chromecast button

domyen:
**Issue** 
Chromecast button shows up on inline videos.
image

**Expected behavior**
Chromecast button should not show up on videos


domyen:
Closing this – I don't have a way to confirm this fix as an iOS user.

PR #58, opened 2019-08-02
Add storybook-addon-responsive-views to list of community addons

juliamitchelmore:
This Storybook addon lets you preview your stories at a variety of breakpoints, so that you can be sure that your components will look great no matter what screen size


domyen:
@juliamitchelmore thanks!

PR #59, opened 2019-08-08
fix: addon alt text

jackdclark:
Use the correct alternative text for the official Console & Links addon images.


domyen:
Thank you @jackdclark! Very much appreciated 🙏

PR #61, opened 2019-08-20
Add GitHub link to header

ghengeveld:
This adds the Octocat image to the top-right corner of the page, linking to Storybook on GitHub. Similar to the docs site.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "email": "dom@hichroma.com"
   },
   "dependencies": {
-    "@storybook/design-system": "^0.0.41",
+    "@storybook/design-system": "^0.0.44",
     "babel-plugin-styled-components": "^1.9.4",
     "gatsby": "^2.3.22",
     "gatsby-image": "^2.0.15",
--- a/site-metadata.js
+++ b/site-metadata.js
@@ -97,14 +97,18 @@ const siteMetadata = {
   },
 };
 
+siteMetadata.urls.navCommunityLinks = [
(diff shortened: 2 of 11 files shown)


ghengeveld:
Schermafbeelding 2019-08-20 om 14 00 57

domyen:
Thanks @ghengeveld I agree with adding a GH link to the header. But I'm not inclined to merge this till we can re-org this section. There's a larger layout challenge of having a lot of top level links. 

My strawman would be to add a dropdown "Ecosystem" which contains Community, Addons, Use cases, Support, GitHub link similar to Vuejs.org.

We already use a dropdown for the mobile view. We can use the same tooltip component in desktop.
 
image

What do you think?

ghengeveld:
The whole point of adding this is to have a direct link to GitHub above the fold. Wanting to go to GitHub from the homepage is a very common thing. I for one always end up on the homepage while I wanted to go to the GitHub page, so a direct click-through is very helpful. Hiding it in a submenu defeats the purpose.

Nevertheless, I suppose we could group some of the other links.

domyen:
How about something like this:
image

ghengeveld:
That would work, but I think it's more obtrusive than the corner banner. All the sudden it's a CTA like the other 2, but it really isn't a CTA. Besides, people know to look for the GitHub link in the top-right corner.

domyen:
I don't think the "dogear" tab makes sense or is universal for open source projects. IIRC, it's a pattern from an old github pages theme. I agree that GitHub  should have more prominent placement, happy to consolidate the header or add a button if that makes sense to you.

PR #69, opened 2019-10-08
Upgrade design system, remove GlobalStyle duplication

kylesuss:
* Upgrade design system to 1.1.6. Mostly features changes to the `Highlight` component which does't appear to be used here, but there are some global styles that were moved around as well.
* Use the `GlobalStyle` from the design system rather than duplicating it here.


kylesuss:
@domyen just did a pass on the deploy preview. LGTM, but want to give it a quick glance plz?

PR #74, opened 2019-11-18
Add the XD addon

KyleW:
There's a plugin for Adobe XD that is similar to the Figma plugin. Adding it to the list in the docs.


domyen:
Awesome thank you

PR #79, opened 2019-12-08
Add storybook-formik addon to list

bhishp:


domyen:
Thanks @bhishp, do you think this would be better in the "Date and State" section rather than "Code" section?

bhishp:
Hi @domyen I wasn't sure the best place to put it to be honest, but Date & State does probably sound more suitable (especially as it already includes addons for other third parties such as Apollo and GraphQL).

I'll adjust the PR now.

Cheers

PR #91, opened 2020-02-08
Add Playroom addon to addon list

rbardini:
## What I did

Add [Playroom](https://github.com/rbardini/storybook-addon-playroom) addon to the _Community addons_ section in addon screen.


domyen:
This is awesome! Thanks for adding @rbardini 🙏

PR #103, opened 2020-05-23
Add Matrix to test addons

tomoya:
As far as I know, props combinations no longer supported.
Ref https://github.com/evgenykochetkov/react-storybook-addon-props-combinations/issues/28

That's why I created the Matrix addon as an alternative.


domyen:
Thank you for the PR @tomoya. 🙇‍♂️ Your addon looks super cool.

PR #106, opened 2020-07-06
Add a script to extract the docs into a path

tmeasday:
This doesn't actually display the docs yet or show any kind of sidebar. It just literally pulls the files in in a sensible way.


scripts/extract-monorepo-docs.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash -
+
+MONOREPO_CODELOAD_URL='https://codeload.github.com/storybookjs/storybook/tar.gz'
+BRANCH=${1:-6.0-docs}
```
kylesuss, line 4: What does this do? 🤔

scripts/extract-monorepo-docs.sh
```diff
@@ -0,0 +1,14 @@
...
+TAR_NAME="storybook-$BRANCH"
+REPO_SUBDIR='docs/src/pages'
+REPO_DEPTH=4 # number of dirs + 1
+REPO_DIRNAME='src/content'
+
+set -e
+
+mkdir -p $REPO_DIRNAME
+rm -rf $REPO_DIRNAME/*
```
kylesuss, line 13: In the release notes PR, I have a `content` dir that has the release notes files in it. It kinda feels like these should live in the same place (`src/content`), but doing so would mean you are deleting that directory each time this is run. Maybe you can namespace the docs 1 more level? So we would have `src/content/releases` and `src/content/docs` or something. WDYT?

kylesuss: I just moved my `content` dir to the same place as you put this:

https://github.com/storybookjs/frontpage/pull/107/commits/744f6066dc4ba02aa53f146436fb69f851e91ee5

So all the MD files will live in the same place. But things will get wonky unless you put the docs 1 level deeper and only clear that out when this script runs.

scripts/extract-monorepo-docs.sh
tmeasday: Uses the first CLI argument, defaulting to `6.0-docs`.

scripts/extract-monorepo-docs.sh
tmeasday: Yeah I saw that after I did this. I'll reconcile them when I merge it.

Issue #105, opened 2020-06-18
What's new page


kylesuss:
* Setup the hierarchy for the what's new page & the various versions
* Link to the what's new page from the header
* Setup page template/component for each version
* Document how to use/add new version docs
* Allow switching between versions
* Setup logic to alter state when shown in an iframe
* Do not index on Google when in iframe

PR #107, opened 2020-07-07
Release notes

kylesuss:
This PR includes:

* A new system to store release notes in `md` files and dynamically pull them into the app
* Backfilled release notes for previous SB versions (5.0+) taken directly from the SB changelog
* An separate route for release notes pages that are served through an iframe
* A custom 404 page for the iframe routes that shows we don't have any release notes for that version (it only affects the iframe pages)

Closes https://github.com/storybookjs/frontpage/issues/105


tmeasday:
@kylesuss I updated the Circle image but I guess its not quite building

.circleci/config.yml
```diff
@@ -2,7 +2,7 @@ version: 2
 jobs:
   build:
     docker:
-      - image: circleci/node:8.12
+      - image: circleci/node:11
```
kylesuss, line 5: @tmeasday I was only able to get this working w/ Node 11. Some sort of nested dependency issue with Node 12 was sending me into a rabbit hole.

PR #114, opened 2020-07-08
Render docs

tmeasday:
Super minimal rendering of the docs based on @kylesuss' work on rendering releases.

This is throwaway work but enough to get us moving.


tmeasday:
Self merging as this is throwaway

PR #118, opened 2020-07-17
Minor fix in Subtitle

crstnio:
Storybook serves more than the Big Three™ now.


domyen:
LGTM thank you @crstnio 🙇

PR #125, opened 2020-08-03
Remove + from contributors

samrobbins85:
This lists the total number of contributors, so it makes more sense without the plus


domyen:
Thanks for the contribution @samrobbins85. The plus is there to indicate that the contributor count is greater than 1035. So we still want it there.

However it's good that you submitted this PR because you called to attention that the `+` character is in the wrong place. Instead of `+1035` it should read `1035+`. 

Is that something you'd be interested in updating in this PR? This pattern is also present in another location on the TeamScreen.

samrobbins85:
Fixed the homepage, but on the team page it seems to be the correct pattern as it shows that there are 1030 developers in addition to the ones shown

domyen:
Nice! Thank you.

PR #126, opened 2020-08-03
Integrate toc

kylesuss:
This component integrates the TableOfContents from the design system into the frontpage to be used w/ the docs. In addition, a bunch of Gatsby updates have taken place to get linking right. I also had to start using a Gatsby "layout" so that the TableOfContents did not re-render on every page change, which is a thing that happens by default.

Lastly, the "guide" navigation for sequential pages was added as was linking to the md file on GitHub.

I am in a bit of a rush, so I didn't do Typescript as that would slow me down 🙈


kylesuss:
Ugh tests failing as a result of #127. Quite an annoying time for that to stop working. I think it is a bit of a wormhole to fix so prob shouldn't be done in this PR.

src/components/layout/DocsLayout.js
```diff
@@ -0,0 +1,186 @@
...
+
+  return (
+    <>
+      <GlobalStyle />
+      <SocialGraph
+        title="Storybook: UI component explorer for frontend developers"
+        desc="Storybook is an open source tool for developing UI components in isolation for React, Vue, and Angular. It makes building stunning UIs organized and efficient."
+        // url={home}
+        // image={ogImage}
+      />
```
kylesuss, line 127: @domyen Do we have a plan/imagery for the social images?

src/components/layout/DocsLayout.stories.js
```diff
@@ -0,0 +1,41 @@
...
+      ],
+    },
+  ],
+};
+
+storiesOf('Frontpage|layout/DocsLayout', module).add('default', () => (
+  <DocsLayout data={data} pageContext={pageContext}>
+    children
+  </DocsLayout>
+));
```
tmeasday, line 41: Lol `storiesOf` ;)

src/components/screens/DocsScreen/DocsScreen.stories.tsx
```diff
@@ -0,0 +1,31 @@
...
+  .add('with guide link, no description', () => (
+    <DocsScreen
+      data={data}
+      pageContext={{ nextTocItem: { ...nextTocItem, description: undefined } }}
+    />
+  ))
+  .add('with github link', () => <DocsScreen data={data} pageContext={{ tocItem }} />)
+  .add('with github link and guide link', () => (
+    <DocsScreen data={data} pageContext={{ tocItem, nextTocItem }} />
+  ));
```
tmeasday, line 31: More `storiesOf` ;)

tmeasday:
I guess it isn't building because it needs that other PR. Would love to see the chromatic changes ;)

gatsby-node.js
```diff
@@ -125,33 +143,60 @@ exports.createPages = ({ actions, graphql }) => {
...
-            } else {
-              console.log(`Not creating page for '/docs/${prefix}/${page}/'`);
+        const docsPagesSlugs = [];
+        const docsPagesEdgesBySlug = docsPagesEdges.reduce(
+          (acc, edge) => ({
+            ...acc,
+            [edge.node.fields.slug]: edge,
+          }),
+          {}
+        );
```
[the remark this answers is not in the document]
kylesuss: Yeah nice. Like this. Will use in the future!

PR #129, opened 2020-08-04
Migrate to MDX. Start using CodeSnippets

kylesuss:
There are some things failing in this PR still, but the general methodology is working here:

https://deploy-preview-129--storybook-frontpage.netlify.app/docs/get-started/whats-a-story/

The first code snippet on there (the button stories) are being generated w/ the new methodology which I describe below. I'd like to get some feedback from you guys (@jonniebigodes & @tmeasday).

Here is how it works:

1. I converted our Gatsby site to MDX. Rather than query MD files, we can now use MDX and embed React components into our MD (or MDX) files. @jonniebigodes I tried to use the `gatsby-remark-embed-snippets` plugin but it seemed to difficult & messy to get the UI that we want (a tabbed interface that is aware of the state of the current framework). That said, I tried to get close to what you were looking for from a maintenance perspective w/ snippets being defined outside of the main MD files.

2. Inside of the main docs MD file, you can import various snippets & pass them to a React component that will render them & only show the ones that are active (whichever snippets are for the current framework).

3. All of the snippets are defined as MDX as well. Here is an example:

https://github.com/storybookjs/storybook/pull/11807/files#diff-43ceb09b19b1793ac23230eb524fb6d4

Within the snippet, you define metadata about it that helps with knowing when to show it on the frontend.

4. The snippets get passed to the frontend, which is aware (not yet, but will be) of the current framework. From there, we can filter the snippets by framework and only show the ones that need to be shown at the moment in a tab interface (not done yet).

----

RE: the approach:

PROS
* Lots of control over the end-user experience
* We can also have markdown snippets to embed (@jonniebigodes called this out in the doc). [Example](https://github.com/storybookjs/storybook/pull/11807/files#diff-46235dbcd9d52771e098d9391d4b2b88)
* Snippets are kept outside of the markdown so they can be updated in isolation
* Storybook supports MDX, so this should be relatively familiar for some people

CONS
* The API is a bit verbose right now: https://github.com/storybookjs/storybook/pull/11807/files#diff-d15ebafce2f148dd652b12b8b6d83960 . I couldn't figure out a good way to `import *` or similar in MDX so I had to individually import the default export & named metadata export.
* Prettier doesn't seem to work too well w/ MDX & rendering React components like this. Notice the indentation on `CodeSnippets`, although MDX & Prettier work well when you wrap something with a code block but sadly that's not what we need here. Not sure what to do about this one right now.

@jonniebigodes the directory structure I chose is pretty arbitrary. You can really pick whatever you want. We just have to make sure the MDX snippet files have the metadata.


tmeasday:
I think the approach is generally the right one but interested in @jonniebigodes' thoughts.

@kylesuss can you explain why you needed to do the aliasing for `useSiteMetadata`? It's sort of unclear to me.

> The API is a bit verbose right now: https://github.com/storybookjs/storybook/pull/11807/files#diff-d15ebafce2f148dd652b12b8b6d83960 . I couldn't figure out a good way to import * or similar in MDX so I had to individually import the default export & named metadata export.

Would another option be something like:

```js
<CodeSnippets
  snippets={[
     '../snippets/button-story/react.js',
     '../snippets/button-story/react.ts',
     '../snippets/button-story/react.md',
     '../snippets/button-story/angular.js',
  ]}
/>
```

And then do the following:

 - Generate metadata based on file name & extension (possibly we could use special comments at the top of the file if more metadata is needed, but I'm not sure if it is).
 - Import the raw snippet on the gatsby side by resolving the filename relative to the source mdx file.

I'm not sure if I am missing something but this seems a little simpler maybe?

kylesuss:
@tmeasday 

> @kylesuss can you explain why you needed to do the aliasing for useSiteMetadata? It's sort of unclear to me.

It was an end of day "sick of gatsby right now" decision. The stories started throwing this error:

https://www.chromatic.com/snapshot?appId=5be26744d2f6250024a9117d&id=5f29ef98e0f1e10022f92b34

This has come up before, but the fix last time was to set the `NODE_ENV` to prod for Storybook builds so that Gatsby would know to handle the GraphQL queries. That didn't work & I also couldn't figure out why it was happening so I just aliased it to try & get the build to pass. But now there are other issues so I think I will circle back and see if I can do that differently.

> Would another option be something like:
```js
<CodeSnippets
  snippets={[
     '../snippets/button-story/react.js',
     '../snippets/button-story/react.ts',
     '../snippets/button-story/react.md',
     '../snippets/button-story/angular.js',
  ]}
/>
```

I just gave this a shot and so far it doesn't look like I can get it to work. There is some magic that happens when you import the MDX file from within another MDX file (it gets parsed by remark) and I can't seem to replicate that by requiring that component from the `CodeSnippets` component.

> Generate metadata based on file name & extension

We can do this, but the metadata needs to have a stylized version of the "language", so as long as the directory can be something like "TS", "JS", "StoriesOf", etc. then I'd say it could work fine. That might feel out of place but maybe it is fine. That said, this approach requires that I get the string import thing figured out so I'll have to see if I can even get it to work w/ some more experiments.

kylesuss:
Got this to work:

https://github.com/storybookjs/storybook/pull/11807/files#diff-d15ebafce2f148dd652b12b8b6d83960

```mdx
import * as buttonStoryReactJS from "../snippets/button-story/react/js.mdx"
import * as buttonStoryReactTS from "../snippets/button-story/react/ts.mdx"
import * as buttonStoryReactMD from "../snippets/button-story/react/md.mdx"
import * as buttonStoryAngularJS from "../snippets/button-story/angular/js.mdx"

<CodeSnippets
  snippets={[
    buttonStoryReactJS,
    buttonStoryReactTS,
    buttonStoryReactMD,
    buttonStoryAngularJS,
  ]}
/>
```

Turns out, it was prettier that was making that not work for me 🤦 . I had to ignore those statements for Prettier. Going to see what I can do to make Prettier play more nicely w/ this syntax.

kylesuss:
Ok! I think this is a pretty good API:

```js
<CodeSnippets
  paths={[
    'react/button-story.js.mdx',
    'react/button-story.ts.mdx',
    'react/button-story.md.mdx',
    'angular/button-story.js.mdx',
  ]}
/>
```

It sets snippets up the way @jonniebigodes wanted in the doc & the paths are pretty simple (no MDX imports 👍 ). We can get the "syntax" from the filename (`.ts.mdx`) & then I can just map that to a "stylized" version of that syntax.

The reason I was having trouble w/ this earlier is that I was trying to use variables in the `import` statement, but webpack didn't know about the files for that reason and therefore they weren't apart of the bundle. I found a workaround so it is working now.

.storybook/main.js
```diff
@@ -42,6 +43,20 @@ module.exports = {
...
+    });
+
+    // TODO: Figure out why Gatsby is throwing this error:
+    // 'The result of this StaticQuery could not be fetched & remove this alias.
+    ['.', '..', '../..', '../../..'].forEach((pathPrefix) => {
+      config.resolve.alias = {
+        ...config.resolve.alias,
+        [`${pathPrefix}/lib/useSiteMetadata`]: p.resolve(__dirname, './useSiteMetadata.js'),
+      };
+    });
```
kylesuss, line 58: This is a hack for sure, but I don't really want to fiddle around with Gatsby anymore right now. It is slowing me down. This feels like an ok solution for the moment given it doesn't affect the primary app. If everyone is ok w/ it, I will file an issue for this to look into post-launch.

package.json
```diff
@@ -7,13 +7,17 @@
     "email": "dom@hichroma.com"
   },
   "dependencies": {
-    "@storybook/design-system": "^5.1.10",
-    "core-js": "^3.6.5",
-    "gatsby": "^2.23.22",
+    "@mdx-js/mdx": "^1.6.16",
+    "@mdx-js/react": "^1.6.16",
+    "@storybook/design-system": "^5.1.11",
+    "core-js": "^2.6.11",
```
kylesuss, line 13: Another thing that feels wrong, but was a time suck to figure out. Storybook was throwing all kinds of core-js errors if I didn't do this. Seems to be a recurring problem given:

https://github.com/storybookjs/frontpage/commit/5402a16b1faeefd654138350467f1df4a155d508#diff-b9cfc7f2cdf78a7f4b91a753d10865a2

.storybook/main.js
```diff
@@ -42,6 +43,15 @@ module.exports = {
...
     ];
 
+    // TODO: Figure out why Gatsby is throwing this error:
+    // 'The result of this StaticQuery could not be fetched' & remove this alias.
+    ['.', '..', '../..', '../../..'].forEach((pathPrefix) => {
+      config.resolve.alias = {
+        ...config.resolve.alias,
+        [`${pathPrefix}/lib/useSiteMetadata`]: p.resolve(__dirname, './useSiteMetadata.js'),
+      };
+    });
```
kylesuss, line 53: This is a hack for sure, but I don't really want to fiddle around with Gatsby anymore right now. It is slowing me down. This feels like an ok solution for the moment given it doesn't affect the primary app. If everyone is ok w/ it, I will file an issue for this to look into post-launch.

jonniebigodes:
@kylesuss sorry for the late response. I was taking care of some other items. I'm checking the changes as i type this. 

Regarding the`gatsby-remark-embed-snippets`plugin usage, it would a good thing for us to use. But it seems that for our case it would offer more issues and problems in terms for usability and maintenance that we can bargain for. I'm going to take a closer look and report back

src/components/screens/DocsScreen/CodeSnippets.tsx
```diff
@@ -0,0 +1,110 @@
...
+  const activeFrameworkPaths = paths.filter((path) => {
+    const [framework] = path.split('/');
+    // TODO: Get the currently active framework rather than hardcoding 'react'
+    return framework === 'react' || framework === COMMON;
+  });
+
+  useEffect(() => {
+    async function fetchModuleComponents() {
+      const fetchedSnippets = await Promise.all(
+        activeFrameworkPaths.map((path, index) => {
```
tmeasday, line 75: You could make this function `async (path, index)` and not need to use `.then()` at the end.

.storybook/main.js
tmeasday: I think maybe the better way to do this is [`NormalModuleReplacementPlugin`](https://webpack.js.org/plugins/normal-module-replacement-plugin/).

Warning: I am not a webpack expert

kylesuss: Nice. This is what I was trying to do w/ alias, but I didn't realize it would happen w/ another strategy. Updated! It works.

kylesuss:
> I wasn't thinking the imported files were going to be mdx, rather than they were just going to be JS (etc) snippets and you were going to use fs.readFileSync or something (would that work in webpack? IDK).

That's a Node API, right? Not sure I can use that in the browser.

That said, I do like the current method because it means that the imported MDX files are already pre-processed & ready to shove into our syntax highlighting component. I'm inclined to go w/ it for that reason, but happy to revisit if it becomes an issue for some reason.

> I suspect what @jonniebigodes was talking about was that some of our examples themselves were going to be MDX files.

Yes! I did check MD files when doing this, as well as MDX (and updated our design system to suit). It seems like syntax highlighting on MDX is sort of weak, but I was able to get it to work well enough with our Markdwown highlighting. Its not perfect, but seems to do the job.

@jonniebigodes this is important -- what that means is that when you are writing snippets for MDX files, write them like this:

` ```md ` rather than ` ```mdx `

But you will define the filename like: `/framework/thing-snippet.mdx.mdx` so that we can show the user it is "MDX".

There isn't a syntax highlighter for MDX in Prism right now, so we have to make do w/ the MD highlighter for now. An alternative that I have seen is some README's using `jsx` to highlight `MDX`. That isn't perfect either, but if need be, we can switch to that later.

.storybook/main.js
```diff
@@ -43,22 +44,22 @@ module.exports = {
...
         resolveFrom: [coreJsLocationOfRoot, coreJsLocationOfGatsby],
       })
     );
 
+    // TODO: Figure out why Gatsby is throwing this error:
+    // 'The result of this StaticQuery could not be fetched' & remove this alias.
+    config.plugins.unshift(
+      new webpack.NormalModuleReplacementPlugin(
+        /lib\/useSiteMetadata\.js/,
+        '../../../.storybook/useSiteMetadata.js'
```
tmeasday, line 59: Maybe use `require.resolve('./useSiteMetadata.js')` rather than this kind of hard to understand path?

PR #134, opened 2020-08-10
Fix links on homepage to framework docs

kylesuss:
We have these on the homepage: 

Screen Shot 2020-08-09 at 6 28 30 PM

And previously they had dedicated guide pages, but I think we removed those, right? In that case, I set up a redirect for each framework at `/docs/framework/` to the first docs page that we have in that framework. Does that sound like the right use case?

@tmeasday || @domyen  `react-native` isn't in the `frameworks.js` file on the SB side & therefore we don't generate docs pages for that framework now -- is that intentional? Should:

1. it be added there?
2. we remove react native from the homepage since we won't have docs for it?


tmeasday:
I think we should remove it from the homepage - @shilman?

domyen:
I linked to the RN readme on GitHub

kylesuss:
The current state is that react-native is not on the homepage. It felt like the "right" thing to do to loop through all of the frameworks that we define in the site metadata. Now all of the links work so this is mergeable, but we can stuff React Native in there if need be.

PR #135, opened 2020-08-10
Fix Pre implementation in the MDXProvider, use in ReleaseScreen as well

kylesuss:
The old `pre` override was causing our `CodeSnippets` (the DS one) to be rendered twice (once for the initial usage from the docs MD file, and another once a pre tag was rendered underneath that). Strangely, I could not replicate this locally (even with a prod Gatsby build), but it was consistent on the deploy for some reason. With that in mind, I built a solution here to make the `Pre` component aware of where in the DOM stack it is being rendered & changing the output based on that.


scripts/extract-monorepo-docs.sh
```diff
@@ -3,7 +3,7 @@
 # Pull the monorepo docs from the branch specified as the first argument, or `master`.
 
 MONOREPO_CODELOAD_URL='https://codeload.github.com/storybookjs/storybook/tar.gz'
-BRANCH=${1:-6.0-docs}
+BRANCH=${1:-next}
```
kylesuss, line 6: This seems right at the moment, but will it eventually be `master`?

tmeasday: Good question. I think probably the default should be `next` and we should set it specifically when the deploying branch is master of that makes sense

kylesuss:
@tmeasday yeah bit of a mystery to me why things are behaving differently on the deploy 🤔  -- would be good to figure that out.

PR #136, opened 2020-08-10
Docs style refinements

domyen:


src/components/layout/Footer.js
```diff
@@ -339,18 +340,15 @@ export default function Footer({ ...props }) {
...
-                    Angular
-                  </ResourceAction>
-                  <ResourceAction withArrow href={docs.home}>
-                    View more
-                  </ResourceAction>
+                  {coreFrameworks.map((framework) => (
+                    <ResourceAction
+                      withArrow
+                      key={framework}
+                      href={`/docs/${framework}/get-started/introduction`}
```
kylesuss, line 347: There is a redirect that was setup at `/docs/${framework}` that you can use (it goes to the same place right now). This was setup in case the first page ever changes in naming scheme.

src/components/layout/Footer.js
```diff
@@ -339,18 +330,15 @@ export default function Footer({ ...props }) {
...
-                  </ResourceAction>
-                  <ResourceAction withArrow href={framework.vue}>
-                    Vue
-                  </ResourceAction>
-                  <ResourceAction withArrow href={framework.angular}>
-                    Angular
-                  </ResourceAction>
-                  <ResourceAction withArrow href={docs.home}>
-                    View more
-                  </ResourceAction>
```
kylesuss, line 353: Ah good catch. I missed these.

PR #139, opened 2020-08-10
Fix copy paste issue for frameworks with missing snippets

kylesuss:
I noticed that the frameworks w/ missing snippets had the messaging about the snippet being missing in the copied text! I added a slot to separate the snippet from this "pre-snippet" content so that only the snippet is copied.


kylesuss:
Merging so this can hit the deploy preview...

PR #141, opened 2020-08-11
Add "for X" to introduction title.

tmeasday:


tmeasday:
3 is better than my solution for sure.

Re: 1 - What I want is really just "is first page" in the context. That's better right? Should I do that?

I'm not sure how 2. would work.

kylesuss:
@tmeasday re: 2, I forgot we are getting the `title` from the frontmatter, so I think that complicates things a bit. Disregard 2. But I was thinking a similar thing as `CodeSnippets`:

```jsx
<!-- prettier-ignore-start -->

<HeadingWithFramework>
 {({ framework }) => `Introduction to Storybook for ${framework}`}
</HeadingWithFramework>

<!-- prettier-ignore-end -->
```

Prob overkill for a heading & it means we have to check if the frontmatter has a `title` value.

> Re: 1 - What I want is really just isFirstTocItem in the context. That's better right? Should I do that? I don't think it'd be too hard.

Yeah that seems better since you can likely get that info without hardcoding the string value of the title & instead rely on the index in a loop or whatever.

PR #176, opened 2020-10-06
Add storybook-addon-vscode-source

vilvaathibanpb:
Added storybook-addon-vscode-source to the addon list. It is a simple storybook addon that opens the source file of the component directly. In local/dev, it opens vs code and in prod deployment, it opens the file in GitHub repository.

An animated gif of the addon


domyen:
Very cool! Thank you 🙇

Issue #171, opened 2020-09-10
404 link in the release page


Laassari:
In the page: [https://storybook.js.org/releases/6.0](https://storybook.js.org/releases/6.0)

the link to `changelogs` is https://github.com/storybookjs/storybook/blob/next/CHANGELOG when it should be https://github.com/storybookjs/storybook/blob/next/CHANGELOG.md

PS: this problem doesn't exist in https://github.com/storybookjs/frontpage/blob/master/src/content/releases/6.0.md

PR #178, opened 2020-10-09
Chore: Fix for issue #171

jonniebigodes:
With this pull request issue #171 is addressed and can be closed.

What was done:
- Changed the links from markdown to `<a>` tags, preventing 404 errors when the reader clicks the links.

The culprit for this issue is `gatsby-remark-link-rewrite, it works in our favor for the bulk extent of the documentation but for cases like this it will work against us, generating invalid urls, as the ones currently live in the releases page. By changing them to ´<a>´ tags, the plugin will bypass the links and will ignore them allowing them to be correct. 

Feel free to provide feedback

### DIFF
--- a/src/content/releases/5.0.md
+++ b/src/content/releases/5.0.md
@@ -13,4 +13,4 @@ title: 'Storybook 5.0 - March 2019'
 
 ---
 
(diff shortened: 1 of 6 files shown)


kylesuss (approved):
Nice. I think this is a good fix for now. Do you think we can eventually customize the Gatsby plugin to just ignore these files somehow? I was just looking at the config for that and it seems it is all regex based.

jonniebigodes:
We could. I've checked the plugin code during the weekend and we could probably go with a `exclude` option to ignore those files.

PR #182, opened 2020-10-14
Add a redirect for the new blog

winkerVSbecks:


winkerVSbecks:
Tested the redirect locally with netlify-cli. It seems to work. However, I'm not sure why it doesn't work with the preview deploy.

tmeasday (approved):
LGTM

PR #186, opened 2020-11-16
Add Contrast Addon to Community Addons under Design

JonathanKolnik:


domyen:
Hey Jono thanks for the PR. The new addons are appended at the end of each category. Will be revamping this later, can you add yours there for now?

PR #190, opened 2020-11-21
upgrade storybook and add addon outline

winkerVSbecks:


kylesuss (approved):
Cool. Seems good. What does the header min-width stuff fixing?

winkerVSbecks:
@kylesuss there is layout shift when once the github stars load. I had added a fix for it previously but, it needs to be slightly bigger because we scale down the github button.

PR #191, opened 2020-11-24
Add LinkWrapper for Docs, Release in nav

kylesuss:
I noticed that the Docs & Releases links in the nav were reloading the page. Kinda hard to keep track of all these LinkWrappers, but this should fix it for the time being.


winkerVSbecks (approved):
Thanks Kyle

PR #193, opened 2020-11-30
Addon catalog UI

winkerVSbecks:
* AddonsHomeScreen
* AddonsCategoryScreen
* AddonsTagScreen
* Search UI components
* AddonsDetailScreen


src/components/basics/Cardinal.js
```diff
@@ -12,19 +12,19 @@ const Count = styled.div`
   color: ${color.dark};
   display: block;
 
-  ${props =>
+  ${(props) =>
```
kylesuss, line 15: Did the prettier config change? Curious why this happened a bunch

winkerVSbecks: Prettier made this change a while back. Don't think it's a config thing. I use the VS code plugin. It respects the project level config.

kylesuss: Oh yeah hmm. I just pulled down and got the same result. I guess we just haven't touched this file in a while ha.

site-metadata.js
```diff
@@ -115,4 +115,14 @@ siteMetadata.urls.navLinks = [
...
 ];
 
+siteMetadata.urls.addonsLearnLinks = [
+  { title: 'How to install addons', icon: 'book', href: '' },
+  {
+    title: 'Create an addon',
+    icon: 'edit',
+    href: 'https://www.chromatic.com/blog/how-to-build-a-storybook-addon/',
+  },
+  { title: 'Add to the catalog', icon: 'add', href: '' },
```
kylesuss, line 125: No href?

src/components/layout/addons/AddonItem.js
```diff
@@ -0,0 +1,263 @@
...
+const VerifiedBadge = styled(Icon).attrs({ icon: 'verified', block: true })`
+  color: ${(props) => (props.appearance === 'official' ? color.secondary : color.green)};
+  width: 14px;
+  height: 14px;
+  margin-left: ${spacing.padding.small}px;
+  margin-bottom: 2px;
+`;
+VerifiedBadge.propTypes = {
+  appearance: PropTypes.oneOf(['official', 'integrator']),
+};
```
kylesuss, line 20: Is there a benefit to adding 0 to a styled component, especially if they aren't required props?

site-metadata.js
winkerVSbecks: Yea, these are placeholders for now. Just building out the UI first while we wait for the back-end and the docs to be updated.

src/components/layout/addons/AddonsLayout.js
```diff
@@ -0,0 +1,266 @@
...
+  hideSidebar: PropTypes.bool,
+};
+
+const ToCContent = styled.div`
+  /* Hide ToC on mobile, the primary navigation is search */
+  display: none;
+
+  ${(props) =>
+    props.hideToC
+      ? css`
```
kylesuss, line 70: How are you deciding when to use `css` in these? This feels a bit out of place as I don't see it elsewhere.

src/components/layout/addons/AddonsList.js
```diff
@@ -0,0 +1,57 @@
...
+
+const ListWrapper = styled.div`
+  > *:not(:last-child) {
+    margin-bottom: ${spacing.padding.medium}px;
+  }
+`;
+
+export const AddonsList = ({ addonItems, isLoading, ...props }) => {
+  const [visibleCount, setVisibleCount] = useState(6);
+  const items = addonItems.slice(0, visibleCount);
```
kylesuss, line 17: Could memoize this so it doesnt alter the array on each render.

kylesuss (approved):
Just looked at the code. Will leave the design to @domyen . Looks great! Only comments were nitpicky things or questions.

src/components/layout/addons/AddonItem.js
winkerVSbecks: imo you get the same benefit as any other component. Was just doing it out of habit but, happy to adopt whatever convention we use in the codebase.

src/components/layout/addons/AddonsLayout.js
winkerVSbecks: I was doing this blindly but, it seems like you only need to do this when interpolating a function. I'll push an update.

src/components/layout/addons/AddonItem.js
kylesuss: Makes sense, though with other components when you have non-required props, you have to specify a default typically -- that's what I was getting at.

src/components/layout/addons/AddonsLayout.js
kylesuss: Np sounds good. I know in previous iterations of the design system that Dom was consistently using css like this but I have noticed it seems to be the exception that you actually need it. Just trying to find common patterns for us.

src/components/layout/addons/AddonItem.js
winkerVSbecks: Good point, these should actually be required

domyen (changes requested):

src/components/layout/addons/AddonsLayout.js
```diff
@@ -0,0 +1,257 @@
...
+
+  #addons-search {
+    font-size: ${typography.size.s2}px;
+    padding-left: 40px;
+    padding-top: 12px;
+    padding-bottom: 12px;
+  }
+
+  svg {
+    left: 16px;
```
domyen, line 106:
```suggestion
    left: 16px;
    font-size: ${typography.size.s2}px;
```

This tweaks makes the icon size larger

PR #196, opened 2020-12-09
docs(addons): add creevey testing tool

wKich:


domyen:
Your addon is now listed. We swapped out the underlying mechanism for the addon catalog [in this project](https://storybook.js.org/blog/revamping-storybooks-addon-ecosystem/) so this PR is not needed.

PR #237, opened 2021-03-05
fail build if tagPages in missing in response

winkerVSbecks:
@shilman added a stricter check. I'm noticing that the builds do end up failing quite regularly. Even before we hit the tags request. Is there something we can do to up the back-end resources?


shilman (approved):
LGTM! I'll see what's going on on the backend 👍

PR #245, opened 2021-03-24
fix typo : npx -> npm

takaya1992:
Can't install addons with the `npx` command 😢 
I think it's correct to use the `npm` command to install the addons.


winkerVSbecks (approved):
Thank you, good catch!

PR #256, opened 2021-04-19
update landing page copy

winkerVSbecks:


domyen:
Updated to:
- Lead with examples of types of UI that SB helps you build: components + pages
- Removed the "react, vue, angular", figure it is redundant now
- Enumerated the build, document, test features in a separate sentence. 

wdyt?

PR #275, opened 2021-06-14
Chore: Changes to the community addons

jonniebigodes:
With this pull request, the framework support table component is updated to properly show the URLs related to the community addons that were moved out of the monorepo:
-design-assets
- events
- cssresources
- graphql
- queryparams
- google-analytics
- knobs 

Allowing the documentation in [here](https://storybook.js.org/docs/react/api/frameworks-feature-support) to show the correct location of said addons.

What was done:
- Created a simple module that we can build on top of for future addons that might be pushed out of the repo.
- Updated the Component to reflect the changes above.

Originally, I thought about moving this information into the Storybook docs toc.js file, but that would lead to include one additional pull request for that specific portion of information and one here to properly handle it. If we can solve it with only one, why not.

Feel free to provide feedback

### DIFF
--- a/src/components/screens/DocsScreen/FrameworkSupportTable.tsx
+++ b/src/components/screens/DocsScreen/FrameworkSupportTable.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import { Link as GatsbyLink } from 'gatsby';
 import stylizeFramework from '../../../util/stylize-framework';
-import { fetchCommunityAddons } from '../../../util/community-addons';
+import { communityAddons } from '../../../util/community-addons';
 
 export function frameworkSupportsFeature(framework, { supported, unsupported }) {
   return (
@@ -21,8 +21,7 @@ export const FrameworkSupportTable = ({ currentFramework, frameworks, featureGro
       return `${monorepoUrlBase}/${repoPath}`;
     }
     // Default is it is an addon (moved into the community or actively maintained)
-    const communityAddon= fetchCommunityAddons(name);
-    return communityAddon || `${monorepoUrlBase}/addons/${name}`;
+    return communityAddons[name] || `${monorepoUrlBase}/addons/${name}`;
   }
 
   return (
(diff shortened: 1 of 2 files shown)


src/util/community-addons.js
```diff
@@ -0,0 +1,42 @@
+const communityAddons = [
```
winkerVSbecks, line 1: You could simplify this to:

```
const communityAddons = {
  'design-assets': 'https://github.com/storybookjs/addon-design-assets',
  'events': 'https://github.com/storybookjs/addon-events',
  // ...
};
```

That way you don't have to do a find for every addon.

src/components/screens/DocsScreen/FrameworkSupportTable.tsx
```diff
@@ -19,8 +20,9 @@ export const FrameworkSupportTable = ({ currentFramework, frameworks, featureGro
     if (repoPath) {
       return `${monorepoUrlBase}/${repoPath}`;
     }
-    // Default is it is an addon
-    return `${monorepoUrlBase}/addons/${name}`;
+    // Default is it is an addon (moved into the community or actively maintained)
+    const communityAddon= fetchCommunityAddons(name);
```
kylesuss, line 24:
```suggestion
    const communityAddon = fetchCommunityAddons(name);
```

src/util/community-addons.js
jonniebigodes: @winkerVSbecks just pushed a commit to address the feedback. Care to take another look?

src/components/screens/DocsScreen/FrameworkSupportTable.tsx
jonniebigodes: Fixed based on @winkerVSbecks feedback

winkerVSbecks (approved):
LGTM

jonniebigodes:
Thanks for the review @winkerVSbecks and @kylesuss . Going to merge this

PR #281, opened 2021-07-05
Optimize SEO for docs

winkerVSbecks:
### What I did

1. No index the documentation for all frameworks except React
2. Index the documentation homepage for all frameworks (eg: so storybook vue” takes you to that page)

@shilman at one point we were generating the React URL as the canonical  link for all docs pages. Like so:

```
<link rel="canonical" href={`${homepageUrl}${buildPathWithFramework(slug, 'react')}/`} />
```

I think the `robots.txt` approach is cleaner since that won't tag the framework specific documentation homepage with the React canonical url.

### DIFF
--- a/src/components/layout/DocsLayout.js
+++ b/src/components/layout/DocsLayout.js
@@ -183,11 +183,18 @@ function DocsLayout({ children, data, pageContext, ...props }) {
     as: 'span',
   };
 
+  // The React specific docs are treated as canonical except for the
+  // docs home page for all other frameworks.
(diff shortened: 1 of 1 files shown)


domyen:
After discussion, the strawman is now to preserve the `rel="canonical"` tag for all docs page BUT the first docs page for a given framework.

That way existing links to pages on non-react frameworks get transferred up to the react-specific page for SEO purposes.

shilman (approved):
LGTM!

PR #303, opened 2021-10-27
Migrate to Emotion

Andarist:


patches/babel-plugin-emotion+10.2.2.patch
```diff
@@ -0,0 +1,112 @@
+diff --git a/node_modules/babel-plugin-emotion/dist/babel-plugin-emotion.cjs.dev.js b/node_modules/babel-plugin-emotion/dist/babel-plugin-emotion.cjs.dev.js
```
kylegach, line 1: Can you share more about this patch?

Andarist: This basically makes the plugin to handle styled imports from your theming package - instead of the `@emotion/styled` pkg.

Most importantly this makes "components as selectors" to work. In v11 we have a babel plugin option to tell us which imports are just wrapper/proxies to our APIs so that could be just configured in the v11. But since you are stuck at v10 i had to hack this with patch-package in all of your repos

kylegach: Aha. Thanks for explaining! 👍

kylegach (approved):
Thanks, @Andarist! Changes look good to me. 👍

winkerVSbecks:
@kylegach is this good to merge now or should we hold off until the versioning docs branch is merged

kylegach:
I'd like to merge #310 first, because that includes a workflow on push to `master`, which we can test with this one.

I'll handle the necessary updates to this PR.

ndelangen (approved):

ndelangen:
@winkerVSbecks WDYT, can we merge this?

winkerVSbecks:
@ndelangen yea, let me fix the linting errors and then I'll merge it

PR #298, opened 2021-10-06
Docs versioning

kylegach:
### Description

**Add versioning to Storybook docs.**

- Via Netlify's branch deploys, build an instance of the docs portion of the site referencing specific versions from the monorepo.
- Use Netlify's redirects to proxy the versioned URL to the branch build and Netlify's headers to prevent non-latest pages from being indexed by search engines.
- Update URLs throughout to include version (unless latest)
    - Ensure relative links, canonical URLs, and social graph images are not broken
- Add a version selector control to the docs sidebar to allow switching between these versions.
- When not viewing latest docs, add a banner pointing to the latest version.
- Other minor changes
    - Remove version from Header
    - Remove "v" prefix from `latestVersion`
    - Only build addons pages when referencing latest version
    
> **Note**: This first attempt at the feature will only support a single "next" version. Once we decide on a convention for multiple "next" versions in the monorepo, I can extend the code here to support those, too (in a future PR).

### Screenshot

Docs page showing the version in the url, version selector in the sidebar, and version callout above the content
_Note: The messaging in the callout above the content is subject to change._

### How to Test

1. Check Chromatic tests
1. Run the site locally (`yarn start` & `yarn build`)
    - Ensure the correct version of docs is built and you can navigate fully
        - Keep in mind the redirects won't work
1. Run the site locally with a reference branch specified (`BRANCH=next yarn start`, same for `build`)
    - Ensure the correct version of docs is built and you can navigate fully
        - Keep in mind the redirects won't work
1. Ensure that `yarn extract-monorepo-docs` and `yarn link-monorepo-docs` still work, with their various arguments
1. Triple-check anything SEO-related (redirect generation code, canonical URLs, headers), please 😉 

#### TODO

- [ ] Ensure algolia search works w/ and w/o versions
    - Fairly sure it works w/o
    - They provided a path for versioning support, but will wait until we get on their latest beta release
- [ ] Create `release-6-0`, `release-6-1`, and `release-6-2` branches from their `release/*.*` counterparts in monorepo

After merging...
- [ ] Create branches in this repo
    - `release-6-0`, `release-6-1`, `release-6-2`
- [ ] Add branches to Netlify build settings

PR #304, opened 2021-10-28
Docs versioning

kylegach:
>**Note for reviewers**
> 
> This is a simpler approach to #298. Notably, the site now extracts all relevant docs versions and builds once, whereas the previous attempt used version-per-branch builds and redirects. While the previous method had shorter build times, it required a complicated and error-prone release process; this new approach requires _only_ updated `releases` files, everything else is automatic (based on the monorepo). 😎 
> 
> I've taken care to organize the commits such that reviewing commit-by-commit should be clear (and doing so is recommended).

### Description

- Update latestVersion format
    - `'v6.3'` -> `6.3`
    - For easier comparison with other verions
    - Remove latest version from Header
- Add script to extract all monorepo docs versions
    - Update prebuild command to use new script
- Add versioned pages with correct URLs
    - Refactor `createPages` to iterate over each version of docs content
    - Move `coreFrameworks`, `communityFrameworks`, and `featureGroups` out of site-metadata
        - These are no longer static, as they depend on the version
        - Place them on `pageContext`
    - Add version to various URLs
    - Use `pageContext.fullPath` instead of `currentPage.fields.slug`, wherever possible
        - `fullPath` includes version & framework
    - Make stories with ToC more realistic
- Add VersionSelector
    - Extract latest and next versions (full) from docs content
        - Place on siteMetaData
    - Generate structured list of versions
        - Place on pageContext
    - Adjust styles & layout
- Add VersionCTA
    - Render on docs pages when not viewing latest
    - Upgrade @storybook/design-system
- Hide non-latest docs from search engines
    - Add redirect from /docs/next/... to /docs/<next_version>/...
- Add more robust error handing to createAddonsPages
- Update README
    - Add warning about potentially unusual behavior

> **Note**: This first attempt at the feature will only support a single "next" version. Once we decide on a convention for multiple "next" versions in the monorepo, I can extend the code here to support those, too (in a future PR).

### Screenshot

Docs page showing the version in the url, version selector in the sidebar, and version callout above the content

### How to Test

1. Check Chromatic tests
1. Run the site locally (`yarn start` & `yarn build`)
    - Ensure the correct versions of docs are built and you can navigate fully
1. Ensure that `yarn extract-monorepo-docs` and `yarn link-monorepo-docs` still work, with their various arguments
1. Triple-check anything SEO-related (redirect from `/docs/next/...`, canonical URLs, headers), please 😉


kylegach:
The build time is not acceptable. Going back to the approach in #298, with significant improvements.

kylegach:
The build time is not acceptable. Going back to the approach in #298, with significant improvements.

PR #317, opened 2021-11-12
Add manual trigger to `push-all-release-branches` workflow

kylegach:
- Add manual trigger to `push-all-release-branches` workflow
- Update versioning docs


kylegach:
This updates GitHub action workflow and internal docs, neither of which are covered by the PR checks here... so I'm merging as-is.

PR #324, opened 2021-12-27
Update _redirects file

kylegach:
Per changes in:

- https://github.com/storybookjs/storybook/pull/16425
- https://github.com/storybookjs/storybook/pull/17074


kylegach:
While these redirects are technically only correct for version `6.4+` and will break in earlier versions, I opted for this simpler approach, as properly accounting for versions would be a significant effort with small benefits.

domyen:
lgtm thanks!

domyen (approved):

PR #340, opened 2022-02-19
Upgrade to DocSearch V3

kylegach:
🚨  **Do not merge this until the V3 index is good-to-go!**

-----

- Add DocsSearch component & stories

### To test the new experience

I temporarily set the env var to the v3 version, so that this deploy preview would build correctly:

https://6218058115c0420008f66c21--storybook-frontpage.netlify.app/docs/vue/get-started/introduction

_Subsequent updates to this PR may affect the style of the search "input" or the autocomplete modal, but the index we can now use through this preview will not be affected by those updates._

### Notable differences from previous implementation

- In general, this new approach is way more "just let Algolia do it" than before
    - We now render their React component, which renders a button which triggers a modal
    - The search "input" is now just a `button` styled to look like an input
        - Which I've further styled to match our Input
            - If you're curious, you can see the before-my-overrides styles in action [here, in the top-right corner](https://docsearch.algolia.com/).
- Icon in the search "input" is slightly different (diffed in Chromatic)
- Cursor is a pointer on hover, not the text caret as before
- Clicking the search "input" opens an Algolia-styled modal with an autocomplete input pre-focused, instead of the previous Storybook-styled dropdown menu
    - I've lightly styled the modal, mostly to use our blue color


src/components/screens/DocsScreen/DocsSearch.tsx
```diff
@@ -0,0 +1,104 @@
...
+    display: none;
+  }
+
+  .DocSearch-Modal {
+    --docsearch-primary-color: ${styles.color.secondary};
+  }
+`;
+const DocSearchStyles = () => <Global styles={docSearchStyles} />;
+
+const label = 'Search docs';
```
winkerVSbecks, line 72: You could mention the keyboard shortcut here:
 
```suggestion
const label = 'Search docs (cmd +k)';
```

kylegach: Good call. Maybe in a tooltip on hover/focus?

winkerVSbecks: Or even just in the placeholder 

CleanShot 2022-02-22 at 19 25 15@2x

domyen: We have a style for the "keyboard shortcut key" in Storybook that looks like this: 
image

image


I believe there's a component for this too!

kylegach: I'll look into it!

kylegach: Boom:

image

There was no shared component, but I found the one used in Storybook's UI and copied it.

----

As an aside, I'm not sure how much sense it makes to use <kbd>^</kbd> to represent <kbd>Control</kbd> for non-Mac users. As I understand it, that's only a Mac convention. That translation comes from Storybook, though, so I kept it as-is to be consistent, and if we want to change it, we should do so over there.

domyen: looks good to me 💪

PR #344, opened 2022-02-24
Update generated `sitemap-all.xml`

kylegach:
- Extract `docs/toc.js` from each docs version
- Use those TOCs to generate a sitemap that only includes pages that are present within each version

-----

You can see the generated file here: https://deploy-preview-344--storybook-frontpage.netlify.app/sitemap-all.xml

Note the pages it contains, for a slug like `/sharing/publish-storybook`, which is present in the "latest" and "next" TOC, but not any others (at the moment):

- ✅  https://storybook.js.org/docs/react/sharing/publish-storybook
- ✅  https://storybook.js.org/docs/6.5/react/sharing/publish-storybook
- ❌  https://storybook.js.org/docs/6.3/react/sharing/publish-storybook
- ❌  https://storybook.js.org/docs/7.0/react/sharing/publish-storybook

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -3,7 +3,7 @@ const { global } = require('@storybook/design-system');
 const siteMetadata = require('./site-metadata');
 const getReleaseBranchUrl = require('./src/util/get-release-branch-url');
 const versionData = require('./src/util/version-data');
-const versions = require('./src/util/versions');
+const { versionsWithToc } = require('./src/util/versions');
 
 require('dotenv').config({
   path: `.env.${process.env.NODE_ENV}`,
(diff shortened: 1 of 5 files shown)


kylegach:
For absolute accuracy, this would also extract `docs/frameworks.js` and only map over the frameworks present for each version, but those frameworks are identical from `6.0` on[^1], so it was avoidable complexity.

[^1]: Except right at this moment, when [`next` has fewer frameworks than the others](https://github.com/storybookjs/storybook/pull/17552), because I haven't propagated that change across `main` and the release branches. But that's a short-lived discrepancy.

gatsby-config.js
```diff
@@ -193,34 +188,47 @@ module.exports = {
...
-                      if (label !== 'latest') {
-                        const parts = page.path.split('/');
-                        parts.splice(2, 0, string);
-                        pagePath = parts.join('/');
-                      }
+                const nonLatestDocsPages = [];
+                versionsWithToc.forEach(({ string, toc }) => {
+                  [...siteMetadata.coreFrameworks, ...siteMetadata.communityFrameworks].forEach(
+                    (framework) => {
+                      const createDocsPages = (tocItems, pathPrefix) => {
```
winkerVSbecks, line 201: There's another `createDocsPages` function in `gatsby-node.js` to actually generate the page. Using the same name here threw me off. Could we rename this to `createDocsPagesInfo` or something.

winkerVSbecks (approved):
Some minor feedback, but otherwise this looks good to me.

PR #353, opened 2022-03-12
Remove extra spacing from top of BlogCTA on desktop

domyen:
Minor one line change to revert previous styling tweak for BlogCTA spacing on desktop.


domyen:
Self-merging because I was responsible for the original spacing tweak which I'm now reverting

PR #358, opened 2022-04-11
Upgrade to Emotion 11

Andarist:


patches/@emotion+babel-preset-css-prop+11.2.0.patch
```diff
@@ -0,0 +1,39 @@
...
++++ b/node_modules/@emotion/babel-preset-css-prop/dist/emotion-babel-preset-css-prop.esm.js
+@@ -25,7 +25,7 @@ var index = (function (api, _temp) {
+   return {
+     plugins: [[pragmatic, {
+       "export": 'jsx',
+-      module: '@emotion/react',
++      module: '@storybook/theming',
+       "import": pragmaName
+     }], [jsx, _extends({
+       pragma: pragmaName,
```
ndelangen, line 39: Do we still need this patch?

ndelangen (changes requested):

patches/@emotion+babel-preset-css-prop+11.2.0.patch
Andarist: This is a different one than before. It piggy-backs on the `@emotion/babel-preset-css-prop` to enable css prop usage without the need to include jsx pragma manually in a file (which nowadays can be configured for the whole project in the React's preset itself, but it requires React 17, and some tools have problem with that setting~). 

So... we could configure this at the `@babel/preset-react` level, although with Gatsby etc this is pretty hidden - but hopefully, there is a way to actually configure this. We could also reevaluate if this is really needed - but I definitely included this patch because without it something wasn't working. It's just that maybe an alternative strategy could be used to get rid of the error.

We could also try to just remove using `gatsby-plugin-emotion` (or however that is called) and "inline" its usage - that would allow us to control this stuff without patches. That gatsby plugin isn't doing a whole lot anyway so... maybe that would be actually for the best?

ndelangen: ok cool!

ndelangen (approved):

patches/@emotion+babel-preset-css-prop+11.2.0.patch
Andarist: Since this got approved I won't be exploring any of those alternative solutions but if you prefer to get rid of this patch - let me know, I'm pretty sure that it's possible. It's just that patching this was the quickest fix to make the project build correctly and I didn't end up resisting this solution.

PR #359, opened 2022-04-12
Add /telemetry redirect

kylegach:
- Add footer link


winkerVSbecks (approved):

site-metadata.js
```diff
@@ -113,6 +114,7 @@ siteMetadata.urls.navCommunityLinks = [
   { title: 'Get involved', href: siteMetadata.urls.community, isGatsby: true },
   { title: 'Use cases', href: siteMetadata.urls.useCases, isGatsby: true },
   { title: 'Support', href: siteMetadata.urls.support, isGatsby: true },
+  { title: 'Telemetry', href: siteMetadata.urls.telemetry, isGatsby: true },
```
kylegach, line 117: Adding this here will add the link to both the Community nav menu in the Header and the Storybook section of the Footer. Preview here: https://deploy-preview-359--storybook-frontpage.netlify.app/

Is that desired? /cc @winkerVSbecks

winkerVSbecks: How about adding it to the footer directly (not via the site metadata). So either it won't show up in tutorials or blog for now or you could patch those footers too. I'm fine with either of those options.

kylegach: Done.

- https://github.com/chromaui/blog/pull/89
- https://github.com/chromaui/learnstorybook.com/pull/596

winkerVSbecks: Thanks!

shilman (commented):
@kylegach @winkerVSbecks Does this need to redirect to the 6.5 docs between now and the time 6.5 goes live?

kylegach:
> @kylegach @winkerVSbecks Does this need to redirect to the 6.5 docs between now and the time 6.5 goes live?

Good point. Imo, yes, it should.

PR #369, opened 2022-06-08
Gatsby upgrade

winkerVSbecks:


src/components/basics/Pre.tsx
```diff
@@ -43,6 +43,7 @@ export function Pre({ children }) {
         />
       );
     }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
kylegach, line 46: The amount of times I'm seeing this rule disabled is a bit worrying. Should we have a follow-up to investigate further?

kylegach (approved):
Good work! One minor comment; no blockers. 🚢 it!

src/components/basics/Pre.tsx
winkerVSbecks: Yea I agree. I do want to fix that but maybe as the next pass. The code didn't change I just updated eslint which started flagging for this.

winkerVSbecks:
@kylegach Going to merge for now and will revisit the exhaustive deps warning

PR #372, opened 2022-06-18
Fix word-break

CindyLyu:
- Fix AddonItem component word-break.

- Before: 
截圖 2022-06-18 下午3 47 54

- After:
截圖 2022-06-18 下午3 51 38


winkerVSbecks (approved):
Thanks for this fix!

PR #381, opened 2022-08-05
add placeholder images

ndelangen:
I am just adding a few placeholder images we can reliably use in storybook for examples purposes without relaying on some external service that might introduce flake.


ndelangen:
Hello @kylegach I'd love to add a few images to the website for placeholder purposes, but the Netlify seems to fail.
Could you help me our?

ndelangen:
I used this to generate these:
https://placeholderimage.dev/

PR #383, opened 2022-08-16
V2

winkerVSbecks:


src/components/screens/DocsScreen/VersionSelector.tsx
```diff
@@ -1,113 +1,59 @@
...
-  line-height: 18px;
-  color: ${color.mediumdark};
-  padding: 7px 15px;
-  border-bottom: 1px solid ${color.border};
-  ${(props) => props.withTopBorder && `border-top: 1px solid ${color.border};`}
-`;
+interface PreRelease {
+  version: number;
+  string: string;
+  label: 'alpha' | 'beta' | 'rc';
```
kylegach, line 19: `7.0` also had a `'future'` label initially, but I'm not sure if that was just a one-time thing. Probably can't hurt to include it.

src/components/screens/DocsScreen/VersionSelector.tsx
```diff
@@ -1,113 +1,59 @@
...
-        <VersionLink isButton primary withArrow>
-          {stylizeVersion(
-            [...versions.stable, ...versions.preRelease].find(({ version: v }) => v === version)
-          )}
-        </VersionLink>
-      </WithTooltip>
-    </Wrapper>
+  const versionOptions = [
+    {
+      label: 'stable',
```
kylegach, line 45: I think these `label`s should be capitalized.

winkerVSbecks: Like 'Stable' instead of `stable`?

kylegach: Yep. Same for "Pre-release".

kylegach: Ha. I left these comments before looking at the snapshots/Storybook. I see now that the visual treatment renders this moot. 🤦‍♂️ Sorry for the noise!

I'll take the opportunity to say that both FrameworkSelector and VersionSelector seem sound and look great to me!

PR #385, opened 2022-09-01
Fix type of `version` prop for DocsSearch

kylegach:
The only reason this hasn't broken the actual experience is because we've passed `versionString` where it's actually used (https://github.com/storybookjs/frontpage/blob/master/src/components/layout/DocsLayout.js#L231). That file isn't in TypeScript, so this type error wasn't caught. 😅


winkerVSbecks:
@kylegach this is going to be replaced by the new nav anyway? Do you still want to merge this in?

winkerVSbecks:
I think I have the types wrong in components-marketing too

kylegach:
Yeah, I realized it was a bit silly to target `master` instead of `v2`. Just changed it.

kylegach:
None of the Chromatic changes look relevant (I don't think it rebuilds when switching the target branch), so I'm going to merge.

PR #389, opened 2022-09-15
disable advanced emotion ssr

winkerVSbecks:


winkerVSbecks:
@Andarist I'm noticing duplicate styles with SSR: https://storybook.js.org
Also react helmet wasn't working correctly. 

I disabled the `gatsby-ssr` stuff you had setup and both issues are resolved. However, I assume you set that up for good reason. Is it safe to remove? Will emotion SSR work without it? Any alternative suggestions?

winkerVSbecks:
Gatsby also recommends using https://www.gatsbyjs.com/docs/reference/config-files/gatsby-ssr/#onRenderBody since it works better with multiple plugins. Could we use that instead?

Andarist:
> @Andarist I'm noticing duplicate styles with SSR: https://storybook.js.org/

I don't see this - which elements should I inspect?

> I disabled the gatsby-ssr stuff you had setup and both issues are resolved. However, I assume you set that up for good reason. Is it safe to remove? Will emotion SSR work without it? Or will it cause other issues (https://github.com/storybookjs/frontpage/pull/321)?

It might cause other issues as mentioned in that linked PR. Ideally, your components wouldn't be written in a way that is incompatible with the default SSR - but I can't guarantee that there are no "offenders" in your whole app ;p

> Gatsby also recommends using https://www.gatsbyjs.com/docs/reference/config-files/gatsby-ssr/#onRenderBody since it works better with multiple plugins. Could we use that instead?

It has been a while since I've used Gatsby - but yet, I don't see any immediate blocker that would prevent you to switch to this method.

winkerVSbecks:
@Andarist you're right the duplicate styles issues is gone. I wonder if updating react helmet helped with that. 

I think I've mostly removed those problematic styles. However, I'm going to try this approach: https://emotion.sh/docs/ssr#gatsby

PR #392, opened 2022-09-16
Fix Storybook demo hero for Safari iOS

ShaunLloyd:
For some reason, the filter property causes Safari mobile to have problems rendering our pointer SVG above the storybook demo images.

using `box-shadow` and `border-radius` to create the shadow of the browser fixes this issue.

### Before
Simulator Screen Recording - iPhone 14 Pro Max - 2022-09-16 at 11 59 57

### After
Simulator Screen Recording - iPhone 14 Pro Max - 2022-09-16 at 12 03 07


winkerVSbecks:
@ShaunLloyd I don't think this works. Since the border radius is a fixed value, it looks strange as you scale down. I'd recommend a % based radius or apply the shadow to a pseudo element. 

CleanShot 2022-09-16 at 12 23 54@2x

PR #393, opened 2022-09-16
use pseudo element dropshadow to fix hero animation on ios 16

winkerVSbecks:


ShaunLloyd (approved):
Awesome! Great catch, @winkerVSbecks 🎉

PR #406, opened 2022-11-01
Add security headers

jmhobbs:
Security headers are a bit lacking, this should tune 'em up!

[Screen Shot 2022-11-01 at 11 20 16 AM](https://securityheaders.com/?q=storybook.js.org&followRedirects=on)

There is some risk here, esp. around `Content-Security-Policy`.  We will want to check the preview, and someone with deep familiarity on the site ensure I didn't miss any third party source.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,3 +1,12 @@
 [build]
   command = "yarn build:with-prefix"
   functions = "/functions"
+[[headers]]
+  for = "/*"
+  [headers.values]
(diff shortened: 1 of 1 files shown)


jmhobbs:
I immediately found more sources in the showcase.  I'll loosen the CSP up for now and we can come back to it.

jmhobbs:
Alternatively we could make it report-only and build up the policy based on reports.

PR #412, opened 2022-11-07
Fix release pages generation

ShaunLloyd:
## What changed?
- Change package.json path for v7.0 onwards to code/package.json


kylegach (approved):
You have better shell-fu than I do. 😅 But I ran it locally and it worked as expected. Thanks, and nice work!

PR #428, opened 2022-11-25
Mock home page for docs contributors

Integrayshaun:
## Why?
Volunteers who help us with our docs can no longer build the site locally because they don't have permission to query the storybookProjects API. `yarn start:doc-only` mocks that call to unblock them.

## What changed
- Add `CreateHomePage` function to mock or build the home page based on `process.env.GATSBY_DOCS_ONLY`


src/components/screens/IndexScreen/IndexScreen.js
```diff
@@ -116,6 +121,7 @@ export function IndexScreen({
 
   return (
     <>
+      <GlobalStyle />
```
winkerVSbecks, line 124: Why is this necessary?

src/pages/index.js
```diff
@@ -1,97 +0,0 @@
...
-      <IndexScreen
-        latestBlogPost={latestPost}
-        projects={projects}
-        npmDownloads={npmDownloads}
-        twitterFollowerCount={twitterFollowerCount}
-        discordMemberCount={discordMemberCount}
-        githubContributorCount={githubContributorCount}
-        youTubeSubscriberCount={youTubeSubscriberCount}
-      />
-    </Fragment>
```
Integrayshaun, line 46: @winkerVSbecks it was being added here in the template file that was deleted.

It's used in all other screen components and we don't get all of our style rules without it :(

src/components/screens/IndexScreen/IndexScreen.js
Integrayshaun: It's needed here now that the index template file has been deleted
https://github.com/storybookjs/frontpage/pull/428#discussion_r1033651834

winkerVSbecks (approved):

PR #429, opened 2022-11-28
Add YouTubeCallout component

kylegach:
- Add stories
- Add to DocsScreen's MdxProvider
- Uses https://www.npmjs.com/package/react-lite-youtube-embed

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,6 +67,7 @@
     "react-dom": "^16.12.0",
     "react-helmet": "^6.1.0",
     "react-lazyload": "^2.6.5",
+    "react-lite-youtube-embed": "^2.3.52",
     "rehype": "^11.0.0",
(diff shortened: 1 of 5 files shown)


domyen:
Awesome, going to commit a few tweaks and this'll be ready to roll

domyen:
Tweaks done! 

Is there a way to `autoplay` the video when you open the summary? I tried adding `params="autoplay=1" muted` per the docs but couldn't get it to work. Perhaps this is a limitation of the library though - no worries if so.

domyen (approved):

kylegach:
> Is there a way to autoplay the video when you open the summary?

I couldn't find a way to do this. I think it's fundamentally opposed to what `lite-youtube-embed` is doing, which is more important, imo.

PR #431, opened 2022-11-29
Tweak styling: Table of contents, VersionCTA

domyen:
Table of Contents: Added more generous spacing between emoji and text to improve readability

image

VersionCTA: adjust spacing to match the YoutubeCallout 
image

### DIFF
--- a/src/components/layout/DocsLayout.js
+++ b/src/components/layout/DocsLayout.js
@@ -79,6 +79,12 @@ const Wrapper = styled.div`
 `;
 
 const StyledTableOfContents = styled(TableOfContents)`
+  // space the ToC emoji and text
+  li > a svg + span::first-letter,
+  li > button svg + span::first-letter {
+    margin-right: 4px;
+  }
+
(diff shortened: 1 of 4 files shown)


src/components/layout/DocsLayout.js
```diff
@@ -79,6 +79,12 @@ const Wrapper = styled.div`
 `;
 
 const StyledTableOfContents = styled(TableOfContents)`
+  // space the ToC emoji and text
```
kylegach, line 82: Do we not want to do this in `@storybook/design-system`?

domyen: I don't think so, the API is to pass an arbitrary string to the ToC component. Sometimes that string might not have an emoji, which will mean the first character is positioned in a strange way.

For example, [the Chromatic docs](https://www.chromatic.com/docs/), which will one-day use the ToC component, don't use emoji.

kylegach (approved):
Thanks for caring about the details!

PR #439, opened 2022-12-06
Update integrations docs

Integrayshaun:
Sibling PR to https://github.com/storybookjs/storybook/pull/20108

### DIFF
--- a/src/components/screens/IntegrationsCatalog/RecipesDetailScreen/RecipesDetailScreen.js
+++ b/src/components/screens/IntegrationsCatalog/RecipesDetailScreen/RecipesDetailScreen.js
@@ -218,7 +218,7 @@ export const RecipesDetailScreen = ({ path, location, pageContext }) => {
               {breadcrumb.title}
             </SubNavBreadcrumb>
             <SubNavRight>
-              <SubNavCTA href="/docs/react/addons/addon-catalog/">
+              <SubNavCTA href="/docs/react/integrations/integration-catalog/">
                 <Icon icon="add" />
                 Add your integration
               </SubNavCTA>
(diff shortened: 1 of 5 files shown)


winkerVSbecks (approved):

src/components/screens/IntegrationsCatalog/RecipesDetailScreen/RecipesDetailScreen.js
```diff
@@ -218,7 +218,7 @@ export const RecipesDetailScreen = ({ path, location, pageContext }) => {
               {breadcrumb.title}
             </SubNavBreadcrumb>
             <SubNavRight>
-              <SubNavCTA href="/docs/react/addons/addon-catalog/">
+              <SubNavCTA href="/docs/react/integrations/integration-catalog/">
```
kylegach, line 221: Should this also be updated? https://github.com/storybookjs/frontpage/blob/main/src/components/screens/IntegrationsCatalog/AddonsDetailScreen/AddonsDetailScreen.js#L195

kylegach (commented):
Changes look good to me!

Because the redirect does not have a wildcard, I think you'll need to add some more specific ones to correspond with these updates, here: https://github.com/storybookjs/frontpage/blob/main/static/_redirects#L113

src/components/screens/IntegrationsCatalog/RecipesDetailScreen/RecipesDetailScreen.js
domyen: good catch KG

domyen (approved):

PR #442, opened 2022-12-06
Redirect to new integrations docs

Integrayshaun:
Fix what I missed in #439 
(Sorry @kylegach)

Also includes updates due to https://github.com/storybookjs/storybook/pull/20186

### DIFF
--- a/src/util/redirects-raw.txt
+++ b/src/util/redirects-raw.txt
@@ -10,7 +10,7 @@
 /docs/workflows/storybook-composition/           /docs/sharing/storybook-composition/                                           301
 /docs/workflows/package-composition/             /docs/sharing/package-composition/                                             301
 /docs/workflows/faq/                             /docs/faq/                                                                     301
-/docs/addons/addon-catalog                       /docs/integrations/integration-catalog                                         301
+/docs/addons/addon-catalog                       /docs/addons/integration-catalog                                               301
 /docs/configure/webpack                          /docs/builders/webpack                                                         301
 /docs/api/addons                                 /docs/addons/introduction                                                      301
 /docs/api/addons-api                             /docs/addons/addons-api                                                        301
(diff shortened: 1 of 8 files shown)


static/_redirects
```diff
@@ -113,6 +113,7 @@
 /addons/                                    /integrations/                                                                                      301
 /recipes/                                   /integrations/                                                                                      301
 /addons/tag/*                               /integrations/tag/:splat                                                                            301
+/docs/react/addons/addon-catalog/           /docs/react/integrations/integration-catalog/                                                       301
```
kylegach, line 116: Whoops. I totally gave you the wrong guidance in #439 🤦‍♂️. `/docs` redirects have some special considerations, due to the dynamic framework and version (or lackthereof). I'll push an update tomorrow, as that's much easier than attempting to explain.

kylegach (changes requested):
I provided the wrong guidance (see comment), so I'm blocking this until I can update it.

src/util/redirects-raw.txt
```diff
@@ -10,6 +10,7 @@
 /docs/workflows/storybook-composition/           /docs/sharing/storybook-composition/                                           301
 /docs/workflows/package-composition/             /docs/sharing/package-composition/                                             301
 /docs/workflows/faq/                             /docs/faq/                                                                     301
+/docs/addons/addon-catalog                       /docs/integrations/integration-catalog                                         301
```
kylegach, line 13: @Integrayshaun — Because of the framework that's part of `/docs/...` URLs, we have to handle those redirects a bit differently. The entries in this file run through this function to generate all of the variations necessary: https://github.com/storybookjs/frontpage/blob/main/gatsby-node.js#L250-L298

kylegach:
Not sure why the PR checks aren't updating, but there's a working deploy preview here: https://deploy-preview-442--storybook-frontpage.netlify.app/docs/react/addons/integration-catalog

PR #453, opened 2022-12-20
Fix LinkWithVersion

kylegach:
- Previously, only added version to hrefs like `../path/to/page`
- Now also adds version to hrefs like `./page`


kylegach:
Not sure why the Netlify checks still show failures, as it deployed successfully (see comment, above).

I was able to build the Storybook locally, so I don't think the Circle CI failure is relevant, either.

Yay, 🤖s!

PR #455, opened 2022-12-21
Recipe edits for styled-components and Emotion

domyen:
**What I did:**
1. Fix typos 
2. Use the spelling of styled-components they use – else we don't get any SEO benefits
3. Add instructions for `GlobalStyle` 

**Why add instructions for `GlobalStyle`?** 
Most apps have globally-scoped CSS, even CSS-in-JS. This is a key part of Storybook's onboarding process that users with existing apps must do to setup successfully. We should document how to apply `GlobalStyle` in Storybook to address this common task.

@kylegach can I get your review on the instructions in styled-components.md? Once you give the green light, I'll apply that to Emotion as well.

### DIFF
--- a/src/content/recipes/styled-components.md
+++ b/src/content/recipes/styled-components.md
@@ -34,7 +34,7 @@ UIs often have a set of global styles that are applied to every component like C
 
 In styled-components, use the [`createGlobalStyle`](https://styled-components.com/docs/api#createglobalstyle) API to scope styles globally instead of locally (which is the library's default behavior).
 
-Open `.storybook/previews.js` and create a `GlobalStyle` that includes a `font-family`. Then apply it to all stories via a [decorator](/docs/react/writing-stories/decorators).
+Open `.storybook/preview.js` and create a `GlobalStyle` component which includes a `font-family`. Then apply it to all stories via a [decorator](/docs/react/writing-stories/decorators).
 
 ```js
 // .storybook/preview.js
--- a/src/content/recipes/styled-components.md
+++ b/src/content/recipes/styled-components.md
@@ -39,6 +39,8 @@ Open `.storybook/preview.js` and create a `GlobalStyle` component which includes
 ```js
 // .storybook/preview.js
 
+import { createGlobalStyle } from 'styled-components';
+
 const GlobalStyle = createGlobalStyle`
   body {
     font-family: "Nunito Sans", "Helvetica Neue", Helvetica, Arial, sans-serif;
@@ -61,7 +63,7 @@ If you already have `GlobalStyle` in your app, you can import it into `.storyboo
 
 </div>
 
-## Using styled-components
+## Use styled-components in components
 
 Let’s update some of our example components to use styled-components instead. Open up the Button component in `./src/stories/button.js.` and replace it with the following code:
 
@@ -228,6 +230,7 @@ Below I created a new file in `.storybook` called `withTheme.decorator.js` that 
 // .storybook/withTheme.decorator.js
 
 import { ThemeProvider } from 'styled-components';
+import { lightTheme } from '../src/theme';
 
 export const withTheme = (Story) => (
   <ThemeProvider theme={lightTheme}>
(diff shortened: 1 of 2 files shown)


src/content/recipes/styled-components.md
```diff
@@ -11,26 +11,59 @@ npx sb init
...
 
-## Using Styled Components
+## How to setup `GlobalStyle`
 
-Let’s update some of our example components to use Styled Components instead. Open up the Button component in `./src/stories/button.js.` and replace it with the following code:
+UIs often have a set of global styles that are applied to every component like CSS resets, `font-size`, `font-family`, and colors.
+
+In styled-components, use the [`createGlobalStyle`](https://styled-components.com/docs/api#createglobalstyle) API to scope styles globally instead of locally (which is the library's default behavior).
+
+Open `.storybook/previews.js` and create a `GlobalStyle` that includes a `font-family`. Then apply it to all stories via a [decorator](/docs/react/writing-stories/decorators).
```
kylegach, line 37:
```suggestion
Open `.storybook/preview.js` and create a `GlobalStyle` component which includes a `font-family`. Then apply it to all stories via a [decorator](/docs/react/writing-stories/decorators).
```

src/content/recipes/styled-components.md
```diff
@@ -11,26 +11,59 @@ npx sb init
...
+    <Story />
+  </>
+);
+
+export const decorators = [withGlobalStyle];
+```
+
+<div class="aside">
+
+If you already have `GlobalStyle` in your app, you can import it into `.storybook/previews.js` instead of creating it anew.
```
kylegach, line 60:
```suggestion
If you already have `GlobalStyle` in your app, you can import it into `.storybook/preview.js` instead of creating it anew.
```

src/content/recipes/styled-components.md
```diff
@@ -220,10 +266,10 @@ export const parameters = {
   },
 };
 
-export const decorators = [withTheme];
+export const decorators = [withGlobalStyle, withTheme];
```
kylegach, line 269: If we flip the order here, you can then access `theme` inside `GlobalStyle`. Perhaps we should demonstrate that, too? Or maybe just a comment explaining why the order matters? Something like this:

```suggestion
// Decorators run in the order they are provided. By placing `withTheme` before `withGlobalStyle`, the GlobalStyle component can access the `theme` via context
export const decorators = [withTheme, withGlobalStyle];
```

src/content/recipes/styled-components.md
```diff
@@ -220,10 +266,10 @@ export const parameters = {
...
 ```
 
-Now, components made with Styled Components will get the theme through the `theme` prop. Let's update the example components to use the theme.
+Now, components made with styled-components will get the theme through the `theme` prop. Let's update the example components to use the theme.
```
kylegach, line 272: Oh, in conjunction with \<previous comment>, we could add an aside after this paragraph, saying that this includes the GlobalStyle component?

kylegach (approved):
Nice edits!

A few minor comments/suggestions, and one not-so-minor one: https://github.com/storybookjs/frontpage/pull/455#discussion_r1055808026

src/content/recipes/styled-components.md
domyen: I think describing this behavior may be too nuanced for a basic recipe, especially given that our `withGlobalStyle` doesn't use `theme`.

src/content/recipes/styled-components.md
domyen: Addressed this [here](https://github.com/storybookjs/frontpage/pull/455/commits/c1fbd29d84c350bac9e519f9c6f09bc635e56c91). What do you think?

src/content/recipes/styled-components.md
kylegach: That's fair. I was concerned about that balance, too. Let's at least not set them up for failure and flip the order?

```suggestion
export const decorators = [withTheme, withGlobalStyle];
```

src/content/recipes/styled-components.md
kylegach: That's a good addition, but not quite what I was referring to here.

src/content/recipes/styled-components.md
Integrayshaun: I thought that they were applied left to right? withGlobalStyles then withTheme would give globaltyles access to the theme

src/content/recipes/styled-components.md
domyen: Chatted to KG in Slack, going to ignore this

PR #460, opened 2023-01-06
Update CodeSnippets logic

kylegach:
- No longer show `common` snippets if non-`common` snippets are available

For this URL, `/docs/web-components/writing-stories/play-function/`, and this instance:

```jsx
<CodeSnippets
  paths={[
    'common/register-component-with-play-function.js.mdx',
    'common/register-component-with-play-function.ts.mdx',
    'web-components/register-component-with-play-function.js.mdx',
    'web-components/register-component-with-play-function.ts.mdx',
  ]}
/>
```

### After

-----

URLs like `/docs/react/writing-stories/play-function/` (i.e. if an exact framework match is unavailable) display the `common` snippets as they did before:

-----

The case where a framework has a match in one language and not another also displays as before. Here, the `react` framework does not match for the `JS` language, so it falls back to `common`. But it _does_ have a match for `TS`, so it uses that directly.

For this URL, `/docs/react/api/csf/`, and this instance:

```jsx
<CodeSnippets
  paths={[
    'common/csf-3-example-starter.js.mdx',
    'web-components/csf-3-example-starter.js.mdx',
    'react/csf-3-example-starter.ts.mdx',
    'vue/csf-3-example-starter.ts-2.ts.mdx',
    'vue/csf-3-example-starter.ts-3.ts.mdx',
    'angular/csf-3-example-starter.ts.mdx',
    'web-components/csf-3-example-starter.ts.mdx',
  ]}
/>
```

> Note: Because there is a `web-components` version of the `JS` language snippet, it would display like the **After** screenshot above.

### DIFF
--- a/src/components/screens/DocsScreen/CodeSnippets.tsx
+++ b/src/components/screens/DocsScreen/CodeSnippets.tsx
@@ -129,16 +129,24 @@ export function CsfMessage({
   );
 }
 
+function filterPathsByFramework(paths: string[], framework: string) {
+  return paths.filter((path) => {
+    const [pathFramework] = path.split('/');
+    return pathFramework === framework;
+  });
+}
+
 export function CodeSnippets({ csf2Path, currentFramework, paths, usesCsf3, ...rest }) {
   const [snippets, setSnippets] = React.useState([]);
-  const activeFrameworkPaths = paths.filter((path) => {
-    const [framework] = path.split('/');
-    return framework === currentFramework || framework === COMMON;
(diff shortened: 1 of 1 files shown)


kylegach:
Chromatic baselines are screwy. Unrelated to this PR.

kylegach:
Merge needs to be coordinated with https://github.com/storybookjs/storybook/pull/20388, because some snippets were missing a language in their filename, which breaks the logic in this PR.

PR #463, opened 2023-01-10
Update recipe and doc layouts for global search indexing

Integrayshaun:
## Why have you done this?
We want to add recipes into the global search which requires some markup changes for Algolia to index them.

## What's changed?
- Use h1 tag for recipe header title
- Remove unused loading state from recipeItemDetail
- Add "Docs > " to hidden algolia indexing title on docs layout
- Change name of directory for sitemap-all

### DIFF
--- a/src/components/layout/DocsLayout.js
+++ b/src/components/layout/DocsLayout.js
@@ -278,7 +278,7 @@ function DocsLayout({ children, isLatest: isLatestProp, pageContext }) {
         <Content>
           {tocSectionTitles && (
             <span hidden id="toc-section-titles">
-              {tocSectionTitles}
+              {`Docs > ${tocSectionTitles}`}
             </span>
           )}
           {(isLatestProp === false || !isLatest) && (
--- a/src/components/layout/DocsLayout.js
+++ b/src/components/layout/DocsLayout.js
@@ -278,7 +278,7 @@ function DocsLayout({ children, isLatest: isLatestProp, pageContext }) {
         <Content>
           {tocSectionTitles && (
             <span hidden id="toc-section-titles">
-              {`Docs > ${tocSectionTitles}`}
+              {`Docs » ${tocSectionTitles}`}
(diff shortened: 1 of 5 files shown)


src/components/layout/DocsLayout.js
```diff
@@ -278,7 +278,7 @@ function DocsLayout({ children, isLatest: isLatestProp, pageContext }) {
         <Content>
           {tocSectionTitles && (
             <span hidden id="toc-section-titles">
-              {tocSectionTitles}
+              {`Docs > ${tocSectionTitles}`}
```
kylegach, line 281: It looks like DocSearch uses "»" as a level seperator:

image

So, we can match that:

```suggestion
              {`Docs » ${tocSectionTitles}`}
```

src/components/layout/integrations/recipes/RecipeItemDetail.js
```diff
@@ -135,11 +104,11 @@ export const RecipeItemDetail = ({
       <RecipeInfo>
         <IntegrationImage icon={icon} accent={accentColor} withConnector />
         <div>
-          <Title isLoading={isLoading}>
-            <span>{isLoading ? 'loading' : `Integrate ${formattedName} and Storybook`}</span>
+          <Title>
+            <span>{`Integrate ${formattedName} and Storybook`}</span>
```
kylegach, line 108: Super minor, but I don't think we need the inner `span`s anymore?

kylegach (approved):
Nice improvements!

We discussed preventing the sidebar on the recipe detail screen from being crawled. Do you need to add a `className` to that markup to be able to do that?

src/components/layout/integrations/recipes/RecipeItemDetail.js
Integrayshaun: You're making a whole lot of sense, Mr. Gach!

PR #483, opened 2023-02-10
Add importance meta to rate docs higher than recipes

Integrayshaun:
Used to tweak cases where recipes are coming up higher than docs when search the name of a framework

The docusearch docs suggest using an integer here so that you can use their sort methods

### DIFF
--- a/src/components/layout/PageLayout.js
+++ b/src/components/layout/PageLayout.js
@@ -121,7 +121,7 @@ export function PurePageLayout({ dxData, children, pageContext, ...props }) {
           <meta
             key={GLOBAL_SEARCH_META_KEYS.IMPORTANCE}
             name={GLOBAL_SEARCH_META_KEYS.IMPORTANCE}
-            content={GLOBAL_SEARCH_IMPORTANCE.RECIPE}
+            content={GLOBAL_SEARCH_IMPORTANCE.AGNOSTIC}
           />
           <link
             rel="preconnect"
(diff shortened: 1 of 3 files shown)


src/components/layout/PageLayout.js
```diff
@@ -107,8 +108,21 @@ export function PurePageLayout({ dxData, children, pageContext, ...props }) {
...
+          <meta
+            key={GLOBAL_SEARCH_META_KEYS.VERSION}
+            name={GLOBAL_SEARCH_META_KEYS.VERSION}
+            content="agnostic"
+          />
+          <meta
+            key={GLOBAL_SEARCH_META_KEYS.IMPORTANCE}
+            name={GLOBAL_SEARCH_META_KEYS.IMPORTANCE}
+            content={GLOBAL_SEARCH_IMPORTANCE.RECIPE}
+          />
```
kylegach, line 125: This applies to anything that _isn't_ docs. I know right now that's only recipes, but it may not always be that way. I think we should either:

1. If we think we'll always want docs content to rate higher than everything, we should change this constant's name to `GLOBAL_SEARCH_IMPORTANCE.AGNOSTIC`
2. If we don't think that, we should apply this important meta only to recipes
    - Either here: https://github.com/storybookjs/frontpage/blob/main/src/components/screens/IntegrationsCatalog/RecipesDetailScreen/RecipesDetailScreen.js
    - Or by updating the logic (and adding a RecipesLayout component) here: https://github.com/storybookjs/frontpage/blob/main/src/components/layout/PageLayout.js

Integrayshaun: Agreed! Let's make an Agnostic option and we can always tweak from there 😀

Integrayshaun: BOOM! Done

kylegach (approved):
Nice improvement!

PR #486, opened 2023-02-15
add event to log sb install

winkerVSbecks:


src/components/screens/DocsScreen/FeatureSnippets.tsx
```diff
@@ -13,14 +20,21 @@ export function PureFeatureSnippets({ framework, snippetsByFramework }) {
...
+    </div>
+  ) : (
     <div className="aside">
       <Snippet />
     </div>
   );
 }
 
 export function FeatureSnippets({ currentFramework, paths }) {
+  const isInstallSnippet = paths[0].includes(INSTALL_PATH);
```
kylegach, line 36: Given that FeatureSnippets is used directly within MDX, explicitly passing an `isInstallSnippet` prop directly to the component seems less finicky to me.

src/components/screens/DocsScreen/FeatureSnippets.tsx
```diff
@@ -13,14 +20,21 @@ export function PureFeatureSnippets({ framework, snippetsByFramework }) {
     }
   }
 
-  return (
+  return isInstallSnippet ? (
+    // eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions
+    <div className="aside" onClick={() => logInstall(framework)}>
```
kylegach, line 25: This approach won't catch people who operate the Copy button with their keyboard. But that's likely done seldomly?

src/components/screens/DocsScreen/FeatureSnippets.tsx
winkerVSbecks: 🤔 will that work for older doc versions too?

kylegach: Alternatively, you could go with a different abstraction and pass the logged event (e.g. `event=""install_storybook"`) to FeatureSnippets in the MDX content.

jonniebigodes: I agree with @kylegach 's reasoning on this.

kylegach: (Sorry, I keep thinking of ideas.)

As a final alternative, you could log something `\`${framework}, ${paths[0]}\`` for _every_ FeatureSnippet (and CodeSnippet?). Then we could map paths to outcomes after we collect the data.

winkerVSbecks:
@kylegach updated to log all snippets

PR #491, opened 2023-02-27
[Recipe] Nextjs framework

Integrayshaun:
## What changed
- Add next.js recipe as a landing page to send people instead of the github readme


src/content/recipes/next.md
```diff
@@ -0,0 +1,84 @@
...
+⤵️ Absolute imports
+🎨 Styling
+🎛 Webpack & Babel config
+💫 and [more](https://github.com/storybookjs/storybook/blob/next/code/frameworks/nextjs/README.md#supported-features)!
+
+## In a project without Storybook
+
+Follow the prompts after running this command in your Next.js project's root directory:
+
+npx storybook@next init
```
kylegach, line 28:
````suggestion
```sh
npx storybook@next init
```
````

File: src/content/recipes/next.md [hunk 2]
```diff
@@ -0,0 +1,84 @@
...
+More on getting started with Storybook
```
kylegach, line 29: Copy/paste error?

src/content/recipes/next.md
```diff
@@ -0,0 +1,84 @@
...
+  // ...
+  addons: [
+    // ...
+    // These can both be removed
+    // 'storybook-addon-next',
+    // 'storybook-addon-next-router',
+  ],
+};
+```
+
```
kylegach, line 79: It's probably a good idea to at least provide an example of how to configure the [router](https://github.com/storybookjs/storybook/blob/next/code/frameworks/nextjs/README.md#nextjs-routing)/[navigation](https://github.com/storybookjs/storybook/blob/next/code/frameworks/nextjs/README.md#nextjs-navigation). Those parts aren't zero-config, because we don't know the shape of their app's routes.

src/content/recipes/next.md
```diff
@@ -0,0 +1,84 @@
...
+</div>
+
+<RecipeHeader>
+
+How to setup Next.js and Storybook
+
+</RecipeHeader>
+
+Next.js is used in tens of thousands of websites and apps. With the recent release of v13, it’s more capable than ever, including many improvements to key features like routing and image optimization. But with all this change, it can be tough to migrate your existing Next app, much less everything it integrates with.
```
kylegach, line 13: Can we run this through ChatGPT to reword? Concerned about SEO with 100% duplicate content.

Integrayshaun: Sure! Good idea

jonniebigodes (approved):
Approving this as to avoid blocking you, as the feedback I have is the same as Kyle

src/content/recipes/next.md
Integrayshaun: How much of the README content would you recommend adding before linking out for more details?

src/content/recipes/next.md
Integrayshaun: Yarp :(

src/content/recipes/next.md
yannbf: Important to note that the api is not 1:1 for the parameters! E.g. For people using storybook-addon-next-router the parameter is nextRouter, same goes for storybook-addon-next, but in @storybook/next it's nextjs.router

kylegach: This recipe doesn't mention the addons at all (aside from the manual migration step to remove them). So not sure we need that nuance?

kylegach (approved):
Looks great!

PR #495, opened 2023-03-03
Create SASS recipe

Integrayshaun:


src/content/recipes/sass.md
```diff
@@ -0,0 +1,60 @@
+<div class="aside aside__no-top">
+
+Some configurations of Storybook already come pre-configured to support sass. If your project meets the following, you're likely ready to go.
```
kylegach, line 3: I can't tell if it's supposed to be `sass` or `Sass`, but we should try to be consistent.

kylegach (approved):
Short & sweet.

src/content/recipes/sass.md
Integrayshaun: Thanks Kyle 😀 Updated

PR #501, opened 2023-03-14
Fall back to TS snippets when TS 4.9 unavailable

kylegach:


kylegach:
Netlify built here https://app.netlify.com/sites/storybook-frontpage/deploys/6410976bc68959007b943a81

Not sure why the status checks aren't reflecting that.

PR #503, opened 2023-03-15
Fix docs anchor links

kylegach:
## What I did

Prior to this change, links to a heading's anchor would often "miss", due to the dynamic snippet loading. This change checks for all snippets to load, then imperatively scrolls to the desired element.

## How to test

1. Open this page of the deploy preview
    - https://deploy-preview-503--storybook-frontpage.netlify.app/docs/react/writing-stories/args#mapping-to-complex-arg-values
2. Confirm that the page doesn't jump around as it loads
    - i.e. That the highlighted target heading stays near the top of viewport


Integrayshaun (approved):
Looks good!

PR #511, opened 2023-03-31
re-tested & fixed headers for release page

kshitijsubedi:
Fixes: https://github.com/storybookjs/storybook/issues/21324

Related: https://github.com/storybookjs/frontpage/pull/508

The last PR didn't fix the iFrame issue. I forgot to maintain the order of the headers config resulting in "/*" overriding the previous headers config.
Also removed obsolete X-Frame-Options and used "frame-ancestors 'self';" instead.
 REF: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options#:~:text=This%20is%20an-,obsolete%20directive,-that%20no%20lonZger

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -15,7 +15,3 @@
   for = "/releases/iframe/*"
   [headers.values]
     Content-Security-Policy = "frame-ancestors *"
-    Referrer-Policy = "strict-origin-when-cross-origin"
-    Permissions-Policy = ""
-    X-Content-Type-Options = "nosniff"
-    X-XSS-Protection = "1; mode=block"
--- a/netlify.toml
+++ b/netlify.toml
@@ -3,20 +3,19 @@
   functions = "/functions"
 
 [[headers]]
-  for = "/releases/iframe/*"
+  for = "/*"
   [headers.values]
-    Content-Security-Policy = "frame-ancestors *"
+    Content-Security-Policy = "default-src 'self'; frame-ancestors 'self'; script-src 'self' 'unsafe-inline' 'unsafe-eval' cdn.segment.io www.google-analytics.com www.googletagmanager.com *.netlify.app data: platform.twitter.com hcaptcha.com *.hcaptcha.com; style-src 'self' 'unsafe-inline' fonts.googleapis.com hcaptcha.com *.hcaptcha.com; img-src * data:; font-src 'self' fonts.gstatic.com; connect-src 'self' *.segment.io www.google-analytics.com *.netlify.app *.algolia.net api.github.com webmention.io hcaptcha.com *.hcaptcha.com storybook.us18.list-manage.com; frame-src app.netlify.com *.widgetbot.io *.youtube.com *.youtube-nocookie.com platform.twitter.com upscri.be hcaptcha.com *.hcaptcha.com *.chromatic.com; media-src * data:;"
     Referrer-Policy = "strict-origin-when-cross-origin"
     Permissions-Policy = ""
     X-Content-Type-Options = "nosniff"
     X-XSS-Protection = "1; mode=block"
 
 [[headers]]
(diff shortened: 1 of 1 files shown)


netlify.toml
```diff
@@ -3,20 +3,19 @@
...
 [[headers]]
-  for = "/*"
+  for = "/releases/iframe/*"
   [headers.values]
-    Content-Security-Policy = "default-src 'self'; script-src 'self' 'unsafe-inline' 'unsafe-eval' cdn.segment.io www.google-analytics.com www.googletagmanager.com *.netlify.app data: platform.twitter.com hcaptcha.com *.hcaptcha.com; style-src 'self' 'unsafe-inline' fonts.googleapis.com hcaptcha.com *.hcaptcha.com; img-src * data:; font-src 'self' fonts.gstatic.com; connect-src 'self' *.segment.io www.google-analytics.com *.netlify.app *.algolia.net api.github.com webmention.io hcaptcha.com *.hcaptcha.com storybook.us18.list-manage.com; frame-src app.netlify.com *.widgetbot.io *.youtube.com *.youtube-nocookie.com platform.twitter.com upscri.be hcaptcha.com *.hcaptcha.com *.chromatic.com; media-src * data:;"
+    Content-Security-Policy = "frame-ancestors *"
     Referrer-Policy = "strict-origin-when-cross-origin"
     Permissions-Policy = ""
     X-Content-Type-Options = "nosniff"
-    X-Frame-Options = "SAMEORIGIN"
```
winkerVSbecks, line 21: This is a requirement. Can you explain why you want to remove it?

kshitijsubedi: During the Test, I found out that the 2nd config i.e /releases/iframe/* inherits headers from the top config of " /* " meaning any unchanged headers are present. so X-frame-Options = SAMEORIGIN was also present in /releases/iframe/* which again blocked the iframe embed.
Since X-Frame-Options is declared Obsolete and using "frame-ancestors" in both Content-Security-Policy worked.
so " /* " have " frame-ancestors self; " equivalent to X-Frame-Options: "SAMEORIGIN" &
"/releases/iframe/*" have " frame-ancestors * " which allows iframe to embed from anywhere.

kshitijsubedi: Just realized I can clean up the config more since /release/iframe/* inherits unchanged headers from /*.

winkerVSbecks: Okay but can you explain where you want to embed these pages? 
We added `X-Frame-Options = "SAMEORIGIN"` after a security audit. I'll need a good reason for removing it.

kshitijsubedi: - to embed on the Release Note page as raised in this issue https://github.com/storybookjs/storybook/issues/21324.
- I am removing x-frame-options and using "frame-ancestors" as a newer method because
	- since x-frame-options only have two options either DENY or SAMEORIGIN, so to allow iframe on any route we have to either remove this header or use frame-ancestor and the issue is if I remove x-frame-options in "/releases/iframe/*" rule then it inherits from first rule i.e "/*" which has x-frame as SAMEORIGIN. 
	[Previously while testing, even though x-frame-options was not in the "/releases/iframe/*" rule it was present as a header inheriting from the above /* rule after deploying since we didn't change that header value]
	- That's why I used Content-Security-Policy = "frame-ancestors" in both rules, that way "/*" having "frame-ancestors self;" will not allow embedding as iframe other same domain but "/releases/iframe/*" only will allow from anywhere with "frame-ancestors *" rule.
	- So technically I am just removing iframe embed from /releases/iframe/* routes only, any other routes will still not allow embedding as an iframe.

winkerVSbecks: Sorry I totally missed that issue. And thanks for the explanation.

winkerVSbecks (approved):

PR #535, opened 2023-04-26
Add docs feedback widget

### LINKED ISSUE #537: Add feedback widget to docs pages

kylegach:
Closes #537 

- Add Feedback component
- Add Feedback to DocsScreen
- Add docs-feedback cloud function
    - Creates/updates discussion on Storybook repo
    - Tracks cumulative rating
    - Re-opens closed discussions with new comments
- Attempt to reduce spam with honeypot and IP rate limiting

> ⚠️ Important note: Even though this is a preview, it is writing to the main Storybook monorepo. So please be careful what you say if you submit a comment and either delete your discussion when you're done or ping me to do it for you.

### Here's what it looks like in-page

Screenshot of feedback widget on the page

### And here's the resulting discussion

Screenshot of discussion on Storybook's repo

### DIFF
--- a/src/components/screens/DocsScreen/Feedback.tsx
+++ b/src/components/screens/DocsScreen/Feedback.tsx
@@ -1,10 +1,12 @@
 import React from 'react';
-import { styled, css } from '@storybook/theming';
-import { Button, Link, styles } from '@storybook/design-system';
+import { styled } from '@storybook/theming';
+import { Button, Link, OutlineCTA, styles } from '@storybook/design-system';
 
 const { code, color, spacing, typography } = styles;
 
 const DOCS_FEEDBACK_URL = '/.netlify/functions/docs-feedback';
+const DISCUSSIONS_URL =
+  'https://github.com/storybookjs/storybook/discussions/categories/documentation-feedback';
 
 const heightTransitionTime = 100; // ms
 
@@ -65,7 +67,7 @@ const CommentForm = styled('form', {
 
   overflow: hidden;
   transition: height ${heightTransitionTime}ms ease-out;
-  height: ${(props) => (props.isExpanded ? `${250 / 16}rem` : 0)};
+  height: ${(props) => (props.isExpanded ? `${260 / 16}rem` : 0)};
 `;
 
 const Label = styled.label`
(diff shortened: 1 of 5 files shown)


src/functions/docs-feedback.js
```diff
@@ -0,0 +1,356 @@
...
+        body: createBody({ isFirst: true, ...received, rating, comment }),
+      },
+    }
+  );
+  console.info('... done!, Added discussion:', '\n', url, '\n', addedTitle, '\n', addedBody);
+
+  return url;
+}
+
+const requestsCache = {};
```
winkerVSbecks, line 293: How does this work? The function gets destroyed after handling the request so, this object will always be empty. No?

winkerVSbecks: Okay so it looks like stuff outside the handler does persist (for an instance) But you could also have multiple instances. And the cache will reset every time the functions is respawned. I guess that's probably fine for now.

https://lihbr.com/posts/rate-limiting-without-overhead-netlify-or-vercel-functions

winkerVSbecks: Definitely a TIL moment

kylegach: Thanks for that link! It verified my approach and taught me that I can get the ip from the function event's headers instead of doing so client-side, which is _much_ simpler.

src/components/screens/DocsScreen/Feedback.tsx
```diff
@@ -0,0 +1,274 @@
...
+  };
+
+  const form = (
+    <CommentForm ref={formRef} isExpanded={expanded} onSubmit={handleSubmit}>
+      {/* eslint-disable-next-line no-nested-ternary */}
+      {error ? (
+        <OutlineCTA badge={<Badge status="error">Error</Badge>} action={<></>}>
+          Something went wrong. Please try again.
+        </OutlineCTA>
+      ) : resultUrl ? (
```
Integrayshaun, line 208: **Nit pick**:

Anyway we can break this down? Nested ternary statements give me the brain scramblies 😵

Integrayshaun (approved):
🚀

PR #542, opened 2023-05-04
Revise feedback format

kylegach:
- Discussion body only contains cumulative rating
- All other feedback in comments

### Before

See #535 

### After

Screenshot of resulting discussion and its comments


Integrayshaun (approved):
🧹 Clean! I like it

PR #545, opened 2023-05-08
Update comment corresponding to #543

kylegach:


kylegach:
This is just a comment, so I'm merging prior to checks passing.

PR #575, opened 2023-07-07
Add IfRenderer component

kylegach:
- Conditionally renders children (newlines before & after required!) based on current renderer

<table>
<tr>
 <th>Before
 <th>After
<tr>
 <td>
 <td>
</table>


jonniebigodes (approved):
Thanks for this @kylegach. Appreciate it 🙏 ! Took it for a spin and it looks good on my end.

PR #586, opened 2023-08-01
Chore: Updates redirects and links to prevent 404s

jonniebigodes:
This pull request follows up on the Storybook documentation pull request [here](https://github.com/storybookjs/storybook/pull/22897) that introduces some structural changes to the documentation organization and outlines the current state of the testing library support.

What was done:
- Updated the redirects
- Updated some misc links across the website to prevent a 404 situation or add yet another redirect to the growing list of them

@kylegach, when you have a moment, can you take a look and follow up with me so that we can merge this one after the already-mentioned documentation pull request is merged?

### DIFF
--- a/src/util/redirects-raw.txt
+++ b/src/util/redirects-raw.txt
@@ -10,7 +10,7 @@
 /docs/writing-docs/doc-block-story               /docs/api/doc-block-story                                                      301
 /docs/writing-docs/doc-block-typeset             /docs/api/doc-block-typeset                                                    301
 /docs/workflows/testing-with-storybook/          /docs/writing-tests/introduction/                                              301
-/docs/workflows/unit-testing/                    /docs/writing-tests/importing-stories-in-tests/                                301
+/docs/workflows/unit-testing/                    /docs/writing-tests/unit-testing/                                              301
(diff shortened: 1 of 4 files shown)


src/util/redirects-raw.txt
```diff
@@ -10,7 +10,8 @@
 /docs/writing-docs/doc-block-story               /docs/api/doc-block-story                                                      301
 /docs/writing-docs/doc-block-typeset             /docs/api/doc-block-typeset                                                    301
 /docs/workflows/testing-with-storybook/          /docs/writing-tests/introduction/                                              301
-/docs/workflows/unit-testing/                    /docs/writing-tests/importing-stories-in-tests/                                301
```
kylegach, line 13: @jonniebigodes — Your original change was:

```diff
- /docs/workflows/unit-testing/  /docs/writing-tests/importing-stories-in-tests/                                301
+ /docs/workflows/unit-testing/  /docs/writing-tests/unit-testing/                                                        301
```

Which would result in the current URL, `/docs/writing-tests/importing-stories-in-tests/` 404'ing.

Instead, we need to have redirects for both the old URL, `/docs/workflows/unit-testing/` and the current one, which both point to the new one.

PR #601, opened 2023-10-02
Recipes: use addon-styling-webpack

Integrayshaun:
## What changed
Update styling recipes to use addon-styling-webpack and addon-themes


src/content/recipes/@vanilla-extract/css.md
```diff
@@ -42,56 +42,19 @@ You can view a full example on [Stackblitz](https://stackblitz.com/edit/github-c
...
-        },
-      ],
-    });
-
-    return config;
-  },
-};
-
-export default config;
+For Webpack users, [@storybook/addon-styling-webpack](https://storybook.js.org/addons/@storybook/addon-styling-webpack/) can get your Storybook up and running with Vanilla Extract in a few lines of code.
```
kylegach, line 45: Is a section like this now appropriate in (some of) the other recipes?

src/content/recipes/less.md
```diff
@@ -16,39 +16,21 @@ How to setup Less and Storybook
 
 Less is a popular CSS preprocessor that allows developers to write more maintainable and reusable stylesheets. Storybook is an industry-standard tool for developing and testing UI components in isolation. With the help of the `@storybook/addon-styling` package, developers can easily incorporate Less stylesheets into their Storybook components.
 
-## Adding `@storybook/addon-styling`
+## Before you begin
 
-<div class="aside aside__no-top">
+If you are using Vite as your storybook builder, all you need to do is import your Less files into your `preview.js` file.
```
kylegach, line 21: storybook → Storybook (saw this a couple other places, as well)

src/content/recipes/@vanilla-extract/css.md
Integrayshaun: Definitely! My only concern is putting this in every relevant recipe before it's tested more thoroughly in the wild

src/content/recipes/less.md
Integrayshaun: Good catch, Gach! Fixed 😁

src/content/recipes/@emotion/styled.md
```diff
@@ -29,44 +29,14 @@ If you’d like to see the example code of this recipe, check out the [example r
...
-### Register the addon
-
-Register the addon with Storybook in `.storybook/main.js`.
-
-```js
-module.exports = {
-  stories: ['../stories/**/*.mdx', '../stories/**/*.stories.@(js|jsx|ts|tsx)'],
-  addons: ['@storybook/addon-essentials', '@storybook/addon-styling'],
-};
+```bash
```
jonniebigodes, line 38: @Integrayshaun one question for you about this pull request. I really liked the fact that you went with the `add` option to install it, but I was wondering, we don’t actually have the option documented, nor the workflow for it (snippets, etc) ,  can users still manually register it or run the codemod to get it set up? If so, we could at least have a callout mentioning it for all the set up instructions. WDYT?

jonniebigodes (approved):
@Integrayshaun left one question for you to follow up on when you have a moment. To unblock you I'm going to approve it.

src/content/recipes/@emotion/styled.md
Integrayshaun: @jonniebigodes I think that's a good call out. What sorts of changes do you think we need to make across the docs to make `storybook add` more widely known?

Users can still manually install and register the addons if they'd like. I just tried to limit the amount of info dumped into the recipes

jonniebigodes: @Integrayshaun I'm working through them and will follow up with you on them.  Sounds good?

PR #603, opened 2023-10-06
Hide v8 from VersionSelector menu

kylegach:
- For now, the v8 docs are only going to be published to provide permalinks. We don't want to expose the option in the menu.


winkerVSbecks (approved):
LGTM. Can I preview the v8 build?

kylegach:
> Can I preview the v8 build?

@winkerVSbecks, thanks for the review! Unfortunately, no. But you can fake it locally:

1. `yarn build`
2. In `src/generated/versions/current/package.json`, change `"version"` to `"8.0.0-future.0"`
3. `yarn start`

PR #604, opened 2023-10-17
Remove trending from catalog homepage

Integrayshaun:
Trending is currently filled with addons that are out of date and unmaintained.
We should hide them for now until we can curate our own list


domyen:
LGTM

kylegach (approved):

PR #644, opened 2023-12-05
Handle snippets with multiple code blocks

kylegach:
### What I did

This fix effectively stitches together a snippet that contains multiple code blocks, like this:

````mdx
```js
// Button.stories.js

import Button from './Button.svelte';
import MarginDecorator from './MarginDecorator.svelte';

export default {
  component: Button,
  decorators: [() => MarginDecorator],
};

// Your stories here.

// Don't forget to use the component you're testing and not the MarginDecorator component
```

```html
<!-- MarginDecorator.svelte -->

<div>
  <slot />
</div>

<style>
  div {
    margin: 3em;
  }
</style>
```
````

Into a unified code block, using the language of the first one:

````mdx
```js
// Button.stories.js

import Button from './Button.svelte';
import MarginDecorator from './MarginDecorator.svelte';

export default {
  component: Button,
  decorators: [() => MarginDecorator],
};

// Your stories here.

// Don't forget to use the component you're testing and not the MarginDecorator component

<!-- MarginDecorator.svelte -->

<div>
  <slot />
</div>

<style>
  div {
    margin: 3em;
  }
</style>
```
````

The end result then displays like:

The syntax highlighting for the second portion is not perfect. This solution compromises that for simplicity of implementation, as there are currently only two examples of multi-code block snippets.

### How to test

There are literally only two snippets that meet the criteria requiring this fix and they're both on the `/writing-stories/decorators` page and apply to the Svelte renderer.

1. Open the [deploy preview's Decorators page](https://deploy-preview-644--storybook-frontpage.netlify.app/docs/writing-stories/decorators)
1. Confirm that the snippets look correct
1. Select the Svelte renderer
1. Confirm that these snippets look correct
    1. Under the [**Wrap stories with extra markup** heading](https://deploy-preview-644--storybook-frontpage.netlify.app/docs/writing-stories/decorators#wrap-stories-with-extra-markup)
    1. Under the [**Component decorators** heading](https://deploy-preview-644--storybook-frontpage.netlify.app/docs/writing-stories/decorators#component-decorators)

### DIFF
--- a/src/components/screens/DocsScreen/CodeSnippets/fetch-snippets.utils.tsx
+++ b/src/components/screens/DocsScreen/CodeSnippets/fetch-snippets.utils.tsx
@@ -64,6 +64,24 @@ const getSnippetTabName = (path: string) => {
   }
   return prettifyName(name);
 };
+
+/**
+ * 1. ModuleComponent is (due to MDX) a React component.
+ * 2. That component renders one or more <pre> tags with a <code> tag inside.
+ * 3. The <code> tag contains the actual code snippet string.
+ */
+function extractContentFromModule(module) {
+  function extractContentFromCodeBlock(block) {
+    return block.props.children.props.children;
+  }
+
+  const containedCodeBlocks = module({ components: {} }).props.children;
+
+  return Array.isArray(containedCodeBlocks)
+    ? containedCodeBlocks.map(extractContentFromCodeBlock).join('\n')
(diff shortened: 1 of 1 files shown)


Integrayshaun:
I think that this is a good way to solve the problem as it is but I think we should re-work this page in the most recent docs to split this into two separate snippets with the IfRenderer

PR #645, opened 2023-12-06
Docs:Next.js Recipe add SWC support

jonniebigodes:
With this small pull request the Next.js recipe was updated to factor in the SWC support recently introduced as part of Storybook 7.6. @valentinpalkovic sorry for not getting this sooner, but it wasn't lost on me. When you or @Integrayshaun have a moment can you take a look and let me know of any feedback you may have. Thanks in advance.

### DIFF
--- a/src/content/recipes/next.md
+++ b/src/content/recipes/next.md
@@ -11,7 +11,6 @@ npx storybook@latest init
 
 ## Set up your project
 
-
 ### In a project without Storybook
 
 Follow the prompts after running this command in your Next.js project's root directory:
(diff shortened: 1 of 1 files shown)


src/content/recipes/next.md
```diff
@@ -69,6 +68,31 @@ export default {
 };
 ```
 
+#### (Experimental) SWC support
+
+If you're working with a Next.js project that already uses SWC (e.g., version 13 or higher), you can enable it in Storybook by adding the following to your `main.js`:
```
valentinpalkovic, line 73: We only support swc mode in Next.js >= 14

jonniebigodes: That's right, thanks for the callout 🙏

valentinpalkovic (approved):

PR #657, opened 2024-01-25
Update package tags script

kylegach:
- Append for `add` command
- Add package disallow list

### DIFF
--- a/scripts/append-package-tags-in-snippets.js
+++ b/scripts/append-package-tags-in-snippets.js
@@ -7,12 +7,18 @@ const { isPreRelease } = versionData;
 
 const DOCS_DIR = path.resolve(__dirname, '../src/content/docs');
 
+const PKG_DISALLOW_LIST = [
+  '@storybook/addon-coverage',
+  '@storybook/addon-webpack5-compiler-babel',
+  '@storybook/addon-webpack5-compiler-swc',
+];
+
 const INLINE_CODE_REGEX = /`(?!`)(.*)`/g;
 const CODE_BLOCK_REGEX = /```(?:sh|shell|bash)\n\s*(?:#.*\n)?(.*)\n\s*```/g;
 
 const INLINE_CODE_MATCH_LIST = (preRelease) => [
   {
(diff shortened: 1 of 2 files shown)


jonniebigodes (changes requested):
@kylegach, thanks for putting this together; however, there's a bit of an issue, and you might have more context. I ran a couple of tests, and it worked out. Nonetheless, after checking the preview deployment, I noticed some issues. Here's what I did:
- Opened up the preview deploy
- Went to the documentation and selected the 8.0 docs
- Started vetting the documentation and noticed that on these pages:
  - [Test coverage](https://deploy-preview-657--storybook-frontpage.netlify.app/docs/8.0/writing-tests/test-coverage)
  - [Install addons](https://deploy-preview-657--storybook-frontpage.netlify.app/docs/8.0/addons/install-addons)
  - [Webpack](https://deploy-preview-657--storybook-frontpage.netlify.app/docs/8.0/builders/webpack)

The next specifier was applied to the snippets, as you can see:
 
coverage-addon-issue

issue-addon-installation-guide

issue-snippet-builders-webpack

This type of issue will become a bit more apparent with the CLI versioning that is being worked on, and we should consider this scenario to avoid misleading the users as the Storybook CLI is showcasing `latest` and the addon version will be the version on `next`. Let me know, and we'll take it from there.

kylegach:
@jonniebigodes —

> - Opened up the preview deploy
> - Went to the documentation and selected the 8.0 docs

The deploy preview only works with the latest version. All of the other versions, like `8.0`, are actually separate sites that are stitched together to appear as one. So as soon as you switched versions, you were no longer on the deploy preview, but the production site (albeit under the deploy preview URL, which is admittedly confusing).

The test suite attempts to ensure that the behavior for both "latest" and "next" builds will be correct.

jonniebigodes (approved):

kylegach:
Follow-up:

- Manipulate `remove` command instances
- Docs repercussions of https://github.com/storybookjs/storybook/issues/25790

PR #658, opened 2024-01-25
Remove conf and jobs board from the header & footer

domyen:
We tried and are now discontinuing these notion docs.


kylegach (approved):
👍

PR #685, opened 2024-03-18
Update URL to 8.0 migration guide

marcusirgens:
ScreenRecording2024-03-18at14 43 09-ezgif com-video-to-gif-converter

Should fix this.


kylegach (approved):
Thanks, @marcusirgens!

PR #711, opened 2025-08-15
add a experimental event-log server-function

ndelangen:
We want to get a better insight into error getting reported.
We want to start reporting the errors to sentry for this reason.

I've implemented a server function that sends telemetry error type events to sentry, and continues to proxy to the google cloud function.


winkerVSbecks:
@ndelangen I think you want to move this to https://github.com/storybookjs/web. This repo is the old version.

